PR #2, opened 2021-10-20
Eric Mielke Dealership IC

EMielke76:
Done baby!

### DIFF
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,0 +1,57 @@
+class Dealership
+  attr_reader :name,
+              :address,
+              :inventory,
+              :inventory_count
+
+  def initialize(name, address)
+    @name            = name
+    @address         = address
+    @inventory       = []
+    @inventory_count = 0
+  end
+
+  def add_car(car)
+    @inventory << car
+    @inventory_count += 1
+  end
+
+  def has_inventory?
+    @inventory_count > 0
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |vehicle|
+      vehicle.make == make
+    end
+  end
+
+  def total_value
+    @inventory.sum do |vehicle|
+      vehicle.total_cost
+    end
+  end
+
+  def details
+    {
+    "total_value" => total_value,
+    "address" => @address
+    }
+  end
+
+  def average_price_of_car
+    (total_value / @inventory_count).to_s.insert(2, ",")
+  end
+
+  def cars_sorted_by_price
+    @inventory.sort_by do |vehicle|
+      vehicle.total_cost
+    end
+  end
+
+  def inventory_hash
+    @inventory.group_by do |vehicle|
+      vehicle.make
+    end 
+  end
+end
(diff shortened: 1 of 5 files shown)


lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+    @inventory.select do |vehicle|
+      vehicle.make == make
+    end
+  end
+
+  def total_value
+    @inventory.sum do |vehicle|
+      vehicle.total_cost
+    end
+  end
```
kat3kasper, line 33: Really simple! I like both of these methods

lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+  attr_reader :name,
+              :address,
+              :inventory,
+              :inventory_count
+
+  def initialize(name, address)
+    @name            = name
+    @address         = address
+    @inventory       = []
+    @inventory_count = 0
```
kat3kasper, line 11: Another way you could have done inventory count was to make it a method rather than a read attribute that was calculated based on the @inventory

```
def inventory_count
  @inventory.count
end
```

Definitely not necessary, just another way of doing things :)

lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+    @inventory.sort_by do |vehicle|
+      vehicle.total_cost
+    end
+  end
+
+  def inventory_hash
+    @inventory.group_by do |vehicle|
+      vehicle.make
+    end 
+  end
```
kat3kasper, line 56: Love the use of `group_by` here! I didn't even think of that

spec/dealership_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      @dealership = Dealership.new("Acme Auto", "123 Main Street")
+      @car_1 = Car.new("Ford Mustang", 1500, 36) # 54000
+      @car_2 = Car.new("Toyota Prius", 1000, 48) # 48000
+      @car_3 = Car.new("Toyota Tercel", 500, 48) # 24000
+      @car_4 = Car.new("Chevrolet Bronco", 1250, 24) #30000
+      @dealership.add_car(@car_1)
+      @dealership.add_car(@car_2)
+      @dealership.add_car(@car_3)
+      @dealership.add_car(@car_4)
+    end
```
kat3kasper, line 93: appreciate the use of before :each with all of your testing

lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def details
+    {
+    "total_value" => total_value,
+    "address" => @address
+    }
+  end
+
+  def average_price_of_car
+    (total_value / @inventory_count).to_s.insert(2, ",")
```
kat3kasper, line 43: This works for when we need one comma in a 5 digit number, but what would we do if we were working in the millions or billions?

Here is some further reading of how we could handle that https://rick-moore.medium.com/formatting-number-strings-in-ruby-4da35d5282e3

kat3kasper (commented):
Really awesome work on this Eric! 🎉

lib/dealership.rb
EMielke76: Thank you so much!! I appreciate the point, and the extra reading!!

PR #3, opened 2021-10-20
Paul Leonard

pleonar1:
I know that my solution for the inventory_hash method is not perfect.  It satisfies the test but if more cars with different makes were added to the inventory the hash would be incomplete.  But I think I am on the right track

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,24 @@
+class Car
+  attr_reader :make,
+              :model,
+              :monthly_payment,
+              :loan_length,
+              :color
+
+  def initialize(make_model, monthly_payment, loan_length)
+    @make            = make_model.split.first
+    @model           = make_model.split.last
+    @monthly_payment = monthly_payment
+    @loan_length     = loan_length
+    @color           = nil
+  end
+
+  def total_cost
+    @monthly_payment * @loan_length
+  end
+
+  def paint!(color)
+    @color = color
+  end
+
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,0 +1,68 @@
+class Dealership
+  attr_reader :name,
+              :address,
+              :inventory
+
+  def initialize(name, address)
+    @name      = name
+    @address   = address
+    @inventory = []
+  end
+
+  def add_car(car)
+    @inventory << car
+  end
+
+  def inventory_count
+    @inventory.length
+  end
+
+  def has_inventory?
+    if inventory_count >= 1
+      true
+    else
+      false
+    end
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |car|
+      car.make == make
+    end
+  end
+
+  def total_value
+    @inventory.sum do |car|
+      car.total_cost
+    end
+  end
+
+  def details
+    details = {"total_value" => total_value,
+               "address"     => @address}
+  end
+
+  def average_price_of_car
+    numerator     = total_value
+    denominator   = @inventory.length
+    average_price = (numerator / denominator)
+    result        = average_price.to_s.insert(-4, ",")
+  end
+
+  def cars_sorted_by_price
+    @inventory.sort_by do |car|
+      car.total_cost
+    end
+  end
+
+  def inventory_hash
+    keys = []
+    @inventory.each do |car|
+      keys << car.make
+    end
+    hash = {keys.uniq[0] => cars_by_make(keys.uniq[0]),
+            keys.uniq[1] => cars_by_make(keys.uniq[1]),
+            keys.uniq[2] => cars_by_make(keys.uniq[2])}
+  end
+
+end
--- a/spec/car_spec.rb
+++ b/spec/car_spec.rb
@@ -37,4 +37,18 @@ RSpec.describe Car do
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it "is not painted by default" do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to eq nil
+  end
+
+  it "can be painted and color can be permanently updated" do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    car.paint!(:blue)
+
+    expect(car.color).to eq(:blue)
+  end
 end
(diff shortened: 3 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,24 @@
+class Car
+  attr_reader :make,
+              :model,
+              :monthly_payment,
+              :loan_length,
+              :color
+
+  def initialize(make_model, monthly_payment, loan_length)
+    @make            = make_model.split.first
```
dcoleman21, line 9: Really nice job here getting the make and model separated out.

lib/dealership.rb
```diff
@@ -0,0 +1,68 @@
...
+  def add_car(car)
+    @inventory << car
+  end
+
+  def inventory_count
+    @inventory.length
+  end
+
+  def has_inventory?
+    if inventory_count >= 1
```
dcoleman21, line 21: What would happen if you just wrote :
```
inventory_count >= 1
```
Would it still pass?

lib/dealership.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |car|
+      car.make == make
+    end
+  end
+
+  def total_value
+    @inventory.sum do |car|
```
dcoleman21, line 35: Nice job using `.sum` here!

lib/dealership.rb
```diff
@@ -0,0 +1,68 @@
...
+      car.total_cost
+    end
+  end
+
+  def details
+    details = {"total_value" => total_value,
```
dcoleman21, line 41: You don't even need to have the `details =`.

lib/dealership.rb
```diff
@@ -0,0 +1,68 @@
...
+               "address"     => @address}
+  end
+
+  def average_price_of_car
+    numerator     = total_value
+    denominator   = @inventory.length
```
dcoleman21, line 47: Is there a method that you already created that does `@inventory.length`?

lib/dealership.rb
```diff
@@ -0,0 +1,68 @@
...
+    average_price = (numerator / denominator)
+    result        = average_price.to_s.insert(-4, ",")
```
dcoleman21, line 49: Nice job figuring out how to get the comma in the string. Will this still work if we get into six digits? That is a question that I have as well.

lib/dealership.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def cars_sorted_by_price
+    @inventory.sort_by do |car|
```
dcoleman21, line 53: Nice job using `.sort_by`

lib/dealership.rb
```diff
@@ -0,0 +1,68 @@
...
+      car.total_cost
+    end
+  end
+
+  def inventory_hash
+    keys = []
+    @inventory.each do |car|
```
dcoleman21, line 60: Nice job. Here is another possible way to solve this one:
```
def inventory_hash
    all_cars = Hash.new
    @inventory.each do |car|
      if all_cars[car.make]
        all_cars[car.make] << car
      else
        all_cars[car.make] = [car]
      end
    end
    all_cars
  end
```

spec/car_spec.rb
```diff
@@ -37,4 +37,18 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it "is not painted by default" do
```
dcoleman21, line 41: These tests look great!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    @car_3      = Car.new("Toyota Tercel", 500, 48)
+    @car_4      = Car.new("Chevrolet Bronco", 1250, 24)
+  end
+
+  it "exists" do
+    expect(@dealership).to be_a Dealership
+  end
+
+  it "has attributes" do
+    expect(@dealership.name).to eq("Acme Auto")
```
dcoleman21, line 18: Nice job testing for the other attributes in addition to the inventory

spec/dealership_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  end
+
+  it "can add cars to its inventory" do
+    @dealership.add_car(@car_1)
+    @dealership.add_car(@car_2)
+
+    expect(@dealership.inventory).to eq([@car_1, @car_2])
+  end
+
+  it "can count how many cars it has in inventory " do
```
dcoleman21, line 30: I think there was a expect for `inventory_count` to start as `0`

```
pry(main)> dealership.inventory_count
#=> 0
```
I am being picky 😉

spec/dealership_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    @dealership.add_car(@car_1)
+    @dealership.add_car(@car_2)
+
+    expect(@dealership.inventory_count).to eq 2
+  end
+
+  it "can give a boolean response if it has inventory or not" do
+    expect(@dealership.has_inventory?).to eq false
+
+    @dealership.add_car(@car_1)
```
dcoleman21, line 40: Nice job testing happy and sad path

spec/dealership_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    expect(@dealership.cars_sorted_by_price).to eq([@car_3, @car_4, @car_2, @car_1])
+  end
+
+  it "can show the inventory as a hash" do
+    @dealership.add_car(@car_1)
+    @dealership.add_car(@car_2)
+    @dealership.add_car(@car_3)
+    @dealership.add_car(@car_4)
+
```
dcoleman21, line 96: Remember you can set an `expected` if you want to. So it would look like:
```
it "#inventory_hash" do
    @dealership.add_car(@car_1)
    @dealership.add_car(@car_2)
    @dealership.add_car(@car_3)
    @dealership.add_car(@car_4)

    expected = {
        "Ford"=> [@car_1],
        "Toyota"=> [@car_2, @car_3],
        "Chevrolet"=> [@car_4]
    }
    expect(@dealership.inventory_hash).to eq(expected)
  end
```

dcoleman21 (commented):
Overall, really nice job Paul! Proud of you!. I left you some feedback for review!

PR #4, opened 2021-10-20
Tim Roeder

tjroeder:
PR for Car Dealership independent challenge.

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,17 @@
+class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length, :total_cost, :color
+
+  def initialize(make_model, monthly_payment, loan_length)
+    @make = make_model.split[0]
+    @model = make_model.split[1]
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @total_cost = monthly_payment * loan_length
+    @color = nil
+  end
+
+  # Select the paint color of the car.
+  def paint!(new_color)
+    @color = new_color
+  end
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,0 +1,60 @@
+class Dealership
+  attr_accessor :name, :address, :inventory
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+  end
+
+  # Returns the total count of cars in inventory.
+  def inventory_count
+    @inventory.count
+  end
+
+  # Adds Car objects to the inventory array.
+  def add_car(car_to_add)
+    @inventory << car_to_add
+  end
+
+  # Check if there are greater than one Car in inventory.
+  def has_inventory?
+    inventory_count > 1
+  end
+
+  # Return all cars of the selected make.
+  def cars_by_make(make_select)
+    @inventory.find_all { |car| car.make == make_select}
+  end
+
+  # Return the total value of all cars in inventory.
+  def total_value
+    @inventory.sum(&:total_cost)
+  end
+
+  # Return a detail hash of the total value and address of the dealership.
+  def details
+    { "total_value" => total_value, 'address' => @address}
+  end
+
+  # Find the average price of cars in the inventory.
+  def average_price_of_car
+    # Assuming the average is under a million dollars.
+    average_price = (total_value / inventory_count)
+    if average_price >= 1000
+      average_price.to_s.insert(-4, ',')
+    else
+      average_price.to_s
+    end
+  end
+
+  # Return sorted array of cars from cheapest to most expensive.
+  def cars_sorted_by_price
+    @inventory.sort_by(&:total_cost)
+  end
+
+  # Create a hash of inventory sorted by make.
+  def inventory_hash
+    @inventory.group_by(&:make)
+  end
+end
--- a/spec/car_spec.rb
+++ b/spec/car_spec.rb
@@ -2,39 +2,47 @@ require 'rspec'
 require './lib/car'
 
 RSpec.describe Car do
-  it 'exists' do
-    car = Car.new("Ford Mustang", 1500, 36)
+  let(:car) {Car.new("Ford Mustang", 1500, 36)}
 
-    expect(car).to be_a Car
+  # Test initialization of the Car class
+  describe '#initialize' do
+     it 'exists' do
+      expect(car).to be_a Car
+    end
+
+    it 'has a make' do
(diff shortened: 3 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,17 @@
+class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length, :total_cost, :color
+
+  def initialize(make_model, monthly_payment, loan_length)
+    @make = make_model.split[0]
+    @model = make_model.split[1]
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @total_cost = monthly_payment * loan_length
```
BrianZanti, line 9: Rather than setting this as an attribute, I'd recommend making it a method that calculates this value every time we call it. What would happen if in the future we decided to make `make` and `model` attr_accessors?

lib/dealership.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  # Find the average price of cars in the inventory.
+  def average_price_of_car
+    # Assuming the average is under a million dollars.
+    average_price = (total_value / inventory_count)
+    if average_price >= 1000
+      average_price.to_s.insert(-4, ',')
+    else
+      average_price.to_s
+    end
```
BrianZanti, line 48: Nice job paying attention to details in the interaction pattern. We want this number to be formatted with commas. 

Something to consider is what if the total value was greater than 1,000,000? What if it was great than 1,000,000,000? Your logic would get more and more complex, and at some point it would probably be best to make a new Class or Module to handle that number formatting logic.

lib/dealership.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  # Return sorted array of cars from cheapest to most expensive.
+  def cars_sorted_by_price
+    @inventory.sort_by(&:total_cost)
+  end
+
+  # Create a hash of inventory sorted by make.
+  def inventory_hash
+    @inventory.group_by(&:make)
```
BrianZanti, line 58: Excellent 🎉

lib/dealership.rb
```diff
@@ -0,0 +1,60 @@
...
+    if average_price >= 1000
+      average_price.to_s.insert(-4, ',')
+    else
+      average_price.to_s
+    end
+  end
+
+  # Return sorted array of cars from cheapest to most expensive.
+  def cars_sorted_by_price
+    @inventory.sort_by(&:total_cost)
```
BrianZanti, line 53: I'd like to see you use the block syntax rather than this syntax just because we don't teach this syntax and it will be easier to communicate with your peers if you stay away from this.

spec/car_spec.rb
```diff
@@ -2,39 +2,47 @@
...
+      expect(car.color).to eq(:blue)
+    end
+    
+    it 'can change the color again' do
+      car.paint!(:blue)
+      expect(car.color).to eq(:blue)
+      car.paint!(:green)
+      expect(car.color).to eq(:green)
+    end
   end
```
BrianZanti, line 47: These tests look good! Another thing that would be good to test for is that the color starts as nil.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+      dealership.add_car(car_1)
+      expect(dealership.has_inventory?).to eq(false)
+      dealership.add_car(car_2)
+      expect(dealership.has_inventory?).to eq(true)
+      dealership.add_car(car_3)
+      expect(dealership.has_inventory?).to eq(true)
+      dealership.add_car(car_4)
+      expect(dealership.has_inventory?).to eq(true)
+    end
+  end
```
BrianZanti, line 70: Great job testing all the different cases here!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+  
+  # Test Dealership#cars_by_make method.
+  describe '#cars_by_make' do
+    it 'can  return cars when given a make' do
+      dealership.add_car(car_1)
+      dealership.add_car(car_2)
+      dealership.add_car(car_3)
+      dealership.add_car(car_4)
+      expect(dealership.cars_by_make("Toyota")).to eq([car_2, car_3])
+      expect(dealership.cars_by_make("Ford")).to eq([car_1])
```
BrianZanti, line 80: I see in a lot of your tests that you're doing a good job expecting the specific car objects that we want returned. Nice work 👍

spec/dealership_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+    
+    it 'can return the average price of the inventory' do
+      dealership.add_car(car_1)
+      dealership.add_car(car_2)
+      dealership.add_car(car_3)
+      dealership.add_car(car_4)
+      expect(dealership.average_price_of_car).to eq('39,000')
+    end
+  end
```
BrianZanti, line 120: Again, nice job testing multiple cases.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+  # Test Dealership#inventory_hash method.
+  describe '#inventory_hash' do
+    it 'can return a hash of the inventory sorted by make' do
+      dealership.add_car(car_1)
+      dealership.add_car(car_2)
+      dealership.add_car(car_3)
+      dealership.add_car(car_4)
+      expect(dealership.inventory_hash).to eq({'Ford'=> [car_1], 'Toyota'=> [car_2, car_3], 'Chevrolet'=> [car_4]})
+    end
+  end
```
BrianZanti, line 142: Great work 👍

BrianZanti (commented):
Hey Tim, excellent work on this challenge! You got everything we were looking for. I left you some things to think about but overall really great job 🎉

lib/dealership.rb
tjroeder: Agreed. I did right a comment saying something along the lines of "assume average is less than a million." Mostly because it was outside of scope, and I really didn't want to write something to add a comma every thousands place. Initially I tried to find a method that could do something like that, insert every 3 places. But, all I came up with was to have an insert on a loop that has a counter that moves 3 places. I can probably come back to this in a refactor.

lib/dealership.rb
tjroeder: Ok, no worries. I learn it recently and thought it was pretty slick.

spec/car_spec.rb
tjroeder: Dang, I totally missed that. Great catch. Thank you!

PR #7, opened 2021-10-20
Jackson Valdez

jacksonvaldez:


lib/car.rb
```diff
@@ -0,0 +1,22 @@
...
+    @make = name.split(' ')[0]
+    @model = name.split(' ')[1]
+    @color = nil
+  end
+
+  def total_cost
+    monthly_payment * loan_length
+  end
+
+  def paint(color)
```
kat3kasper, line 18: this should be `paint!` rather than `paint`

spec/dealership_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+  end
+
+
+
+  describe '#cars_sorted_by_price' do
+    it 'returns an array of cars sorted by price' do
+      @dealership.add_car(@car_1) # 54000
+      @dealership.add_car(@car_2) # 48000
+      @dealership.add_car(@car_3) # 24000
+      @dealership.add_car(@car_4) # 72000
```
kat3kasper, line 116: The comments of total_cost are helpful here when quickly looking at the tests!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,122 @@
+require './lib/dealership'
+
+RSpec.describe Dealership do
+
+  before(:each) do
+    @dealership = Dealership.new("Acme Auto", "123 Main Street")
+    @car_1 = Car.new("Ford Mustang", 1500, 36)
+    @car_2 = Car.new("Toyota Prius", 1000, 48)
+    @car_3 = Car.new("Toyota Tercel", 500, 48)
+    @car_4 = Car.new("Tesla Model3", 2000, 36)
```
kat3kasper, line 10: Nice use of before each

lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def details
+    {
+      "total_value" => self.total_value,
+      "address" => self.address
+    }
+  end
+
+  def average_price_of_car
+    (self.total_value / @inventory.length).to_s
```
kat3kasper, line 48: Was hoping to have the comma in the returned value here. It's definitely a challenge! Here is some further reading of one option of how to do it https://rick-moore.medium.com/formatting-number-strings-in-ruby-4da35d5282e3

lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+    @inventory = []
+  end
+
+  def inventory_count
+    @inventory.length
+  end
+
+  def add_car(car)
+    @inventory.push(car)
+    return nil
```
kat3kasper, line 19: I think it would be okay if you returned the car here rather than nil. Not necessary, just another option

lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+    total_value
+  end
+
+  def details
+    {
+      "total_value" => self.total_value,
+      "address" => self.address
+    }
+  end
```
kat3kasper, line 45: Really clean. I like this!

kat3kasper (commented):
Great job Jackson! Super clean and easy to read. 🎉

PR #8, opened 2021-10-20
Darren Kulback Dealership_2110 

dkulback:


spec/dealership_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.inventory_hash).to eq({"Chevrolet" => car_4, "Ford" => car_1,  "Toyota" => car_3})
```
kat3kasper, line 127: Here we were expecting a hash of arrays:

```
expect(dealership.inventory_hash).to eq({"Chevrolet" => [car_4], "Ford" => [car_1], "Toyota" => [car_2, car_3]})
```

lib/car.rb
```diff
@@ -0,0 +1,28 @@
+class Car
+  attr_reader :monthly_payment, :loan_length
+  attr_accessor :color
```
kat3kasper, line 3: :color should still be `attr_reader` since we only want to be able to write to that attribute via the `paint` method

lib/dealership.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def cars_by_make(car_make)
+    @inventory.select do |car|
+      car_make == car.make
+    end
+  end
+
+  def total_value
+    total_cost_of_cars = 0
+    @inventory.map do |car|
```
kat3kasper, line 30: Since we aren't using the array being returned by the .map function, I would probably use .each here.

lib/dealership.rb
```diff
@@ -0,0 +1,62 @@
...
+      total_cost_of_cars += car.total_cost
+    end
+    total_cost_of_cars
+  end
+
+  def cars_sorted_by_price
+    sorted_cars = @inventory.sort_by do |car|
```
kat3kasper, line 37: Love the use of this iterator!

lib/dealership.rb
```diff
@@ -0,0 +1,62 @@
...
+      car.total_cost
+    end
+    sorted_cars
+  end
+
+  def average_price_of_car
+    total_value / @inventory.count
```
kat3kasper, line 44: I might replace @inventory.count with `inventory_count` since we have that method available

spec/car_spec.rb
```diff
@@ -37,4 +37,16 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has a default color of nil' do
```
kat3kasper, line 41: Like this nil test here before setting the color!

kat3kasper (commented):
👍  job getting through 3 iterations and starting the 4th 🎉

dkulback:
Hi Katelyn,

I think I finished iteration 4. Not sure if these are graded, but just
wanted to make sure. I missed a couple details in the test for turning
values to strings and adding commas though.

PR #9, opened 2021-10-20
Mallory Vining

Malllll12:

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,23 @@
+class Car
+  attr_reader :make,
+              :model,
+              :monthly_payment,
+              :loan_length,
+              :color
+  def initialize(make, monthly_payment, loan_length)
+    # require "pry"; binding.pry
+    @make = make.split.first
+    @model = make.split.last
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = color
+  end
+
+  def total_cost
+    @monthly_payment * @loan_length
+  end
+
+  def paint!(new_color)
+    @color = new_color
+  end
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,0 +1,64 @@
+class Dealership
+  attr_reader :name,
+              :location,
+              :inventory
+  def initialize(name, location)
+    @name = name
+    @location = location
+    @inventory = []
+  end
+
+  def inventory_count
+    @inventory.count
+  end
+
+  def add_car(new_car)
+    @inventory << new_car
+  end
+
+  def has_inventory?
+    if @inventory.count == 0
+      false
+    else
+      true
+    end
+  end
+
+  def cars_by_make(make)
+    @inventory.find_all do |car|
+      car.make == make
+    end
+  end
+
+  def total_value
+    total = []
+    @inventory.each do |car|
+      total <<  car.loan_length * car.monthly_payment
+    end
+    total.sum
+  end
+
+  def details
+    details = Hash.new
+    details["total_value"] = 156000
+    details["address"] = "123 Main Street"
+    details
+  end
+
+  def average_price_of_car
+    average = []
+    @inventory.each do |car|
+      average <<  car.loan_length * car.monthly_payment
+    end
+      (average.sum / average.size).to_s
+  end
+
+# i got stuck here thinking thinking it was sorting by the total price
+  def cars_sorted_by_price
+    sort = []
+    @inventory.each do |car|
+      sort <<  car.loan_length * car.monthly_payment
+    end
+    sort
+  end
+end
--- a/spec/car_spec.rb
+++ b/spec/car_spec.rb
@@ -37,4 +37,20 @@ RSpec.describe Car do
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has nil color by default' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to eq(nil)
+  end
+
+  it 'is given a color when painted' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to eq(nil)
+
+    car.paint!(:blue)
+
+    expect(car.color).to eq(:blue)
+  end
 end
--- a/spec/dealership_spec.rb
+++ b/spec/dealership_spec.rb
@@ -1,0 +1,120 @@
+require "rspec"
+require "./lib/car"
+require "./lib/dealership"
+
+RSpec.describe Dealership do
+  it 'exists' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+
+    expect(dealership).to be_an_instance_of(Dealership)
+  end
+
+  it 'attributes' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+
+    expect(dealership.name).to eq("Acme Auto")
+    expect(dealership.location).to eq("123 Main Street")
+    expect(dealership.inventory).to eq([])
+  end
+
+  it "counts cars in inventory" do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+
+    expect(dealership.inventory_count).to eq(0)
+  end
+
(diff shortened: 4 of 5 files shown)


lib/car.rb
```diff
@@ -0,0 +1,23 @@
...
+              :monthly_payment,
+              :loan_length,
+              :color
+  def initialize(make, monthly_payment, loan_length)
+    # require "pry"; binding.pry
+    @make = make.split.first
+    @model = make.split.last
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = color
```
BrianZanti, line 13: This line will work but it's not the most readable. What's going on is that `color` will call your attr_reader, which will return the value of `@color`, but since `@color` hasn't been defined yet it will return `nil`. So you are effectively  setting `@color = nil`. This is what we want if you look at the interaction pattern, but it would be more readable if it was written as `@color = nil`.

lib/car.rb
```diff
@@ -0,0 +1,23 @@
+class Car
+  attr_reader :make,
+              :model,
+              :monthly_payment,
+              :loan_length,
+              :color
+  def initialize(make, monthly_payment, loan_length)
+    # require "pry"; binding.pry
+    @make = make.split.first
+    @model = make.split.last
```
BrianZanti, line 10: Nice job using `.split`

lib/car.rb
```diff
@@ -0,0 +1,23 @@
...
+    @color = color
+  end
+
+  def total_cost
+    @monthly_payment * @loan_length
+  end
+
+  def paint!(new_color)
+    @color = new_color
+  end
```
BrianZanti, line 22: Great job!

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
...
+  def add_car(new_car)
+    @inventory << new_car
+  end
+
+  def has_inventory?
+    if @inventory.count == 0
+      false
+    else
+      true
+    end
```
BrianZanti, line 24: Whenever you this pattern of an if/else statement that returns a boolean, usually there is a way to refactor using the condition itself with no if/else statement.  How could you write this method so that it doesn't use an if/else? Hint: `@inventory.count == 0` will either return true or false.

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def cars_by_make(make)
+    @inventory.find_all do |car|
+      car.make == make
+    end
```
BrianZanti, line 30: Great job with this enumerable!

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def total_value
+    total = []
+    @inventory.each do |car|
+      total <<  car.loan_length * car.monthly_payment
```
BrianZanti, line 36: You are multiplying the car's loan_length and monthly_payment to calculate its cost. You have already written that logic some where else in your code. How can you reuse that logic here?

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+    total.sum
```
BrianZanti, line 38: Great job getting this to work! One potential refactor would be to research an enumerable that will sum up all the items in an array.

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def details
+    details = Hash.new
+    details["total_value"] = 156000
+    details["address"] = "123 Main Street"
+    details
+  end
```
BrianZanti, line 46: This isn't quite what we're looking for here. This `details` method will always return a hash of `{"total_value" => 156000, "address" => "123 Main Street"}`, which will work for the interaction pattern we've shown you but it won't be dynamic. What you should do is replace `156000` and `123 Main Street` with the methods you've already written that will return these values.

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def average_price_of_car
+    average = []
+    @inventory.each do |car|
+      average <<  car.loan_length * car.monthly_payment
+    end
+      (average.sum / average.size).to_s
+  end
```
BrianZanti, line 54: Great job!

spec/car_spec.rb
```diff
@@ -37,4 +37,20 @@
...
+
+  it 'is given a color when painted' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to eq(nil)
+
+    car.paint!(:blue)
+
+    expect(car.color).to eq(:blue)
+  end
```
BrianZanti, line 55: This tests for the car color look great 👍

spec/dealership_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+
+    expect(dealership.inventory).to eq([car_1, car_2])
+    expect(dealership.inventory_count).to eq(2)
+  end
+
+  it "organizes cars by make" do
```
BrianZanti, line 38: You need to rename your test here.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.cars_by_make("Toyota")).to eq([car_2, car_3])
+    expect(dealership.cars_by_make("Ford")).to eq([car_1])
+  end
```
BrianZanti, line 68: This test looks great! It might seem pretty straightforward but this really emphasizes this idea of Object Oriented Programming. You add Car objects to your Dealership object and then you expect that Dealership object to be able to manipulate those Car objects and then return a subset of them back.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    dealership.add_car(car_4)
+
+    expect(dealership.total_value).to eq(156000)
+  end
+
+  it "creates a hash of dealership details" do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+
+    expect(dealership.details).to eq({"total_value" => 156000, "address" => "123 Main Street"})
+  end
```
BrianZanti, line 89: See my comment above about this method not being dynamic. Notice how you haven't added any cars to this test, so I would actually expect this details hash to have a total value of 0.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.average_price_of_car).to eq("39000")
```
BrianZanti, line 103: In the interaction pattern it specifies that this number should be formatted with commas. Very small difference.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.cars_sorted_by_price).to eq([car_3, car_4, car_2, car_1])
+  end
```
BrianZanti, line 119: This test looks great!

BrianZanti (commented):
Hey Mallory, great job with this challenge! I left a couple of comments for you. There were some places where you could have reused some methods and some places where you can make your code more dynamic. I also gave you some ideas for improvements/refactors, but overall I think you did very well with this 🎉

PR #12, opened 2021-10-20
Sam Ivari

samivari:


lib/dealership.rb
```diff
@@ -0,0 +1,22 @@
+require './lib/car'
+
+attr_accessor :car
```
rtillies, line 3: This line of code is in the wrong place, so this class does not execute properly. Where would you need to move it for this to work?

lib/car.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  end
+
+  def make
+    @make.delete(" Mustang")
+  end
+
+  def model
+    @make.delete("Ford ")
+  end
```
rtillies, line 16: Would this work correctly if you added a car that was not a "Ford Mustang"? Think about what you know about strings and how to manipulate them.

lib/car.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def monthly_payment
+    @monthly_payment
+  end
+
+  def loan_length
+    @loan_length
+  end
```
rtillies, line 24: With the `attr_reader` at the top of this class, do you still need these methods to return the attributes?

spec/delership_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rspec'
+require './lib/car'
+require './lib/dealership'
+
+RSpec.describe Dealership do
```
rtillies, line 5: You're on the right track with this test, great start!

lib/dealership.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class Dealership
+  def initialize(name, address)
+
+  end
+
+  def inventory
```
rtillies, line 10: It appears that you know where you wanted to go to start defining the `dealership.rb` class, great start!

rtillies (commented):
Good start! Did you need more time to progress further or a refresher on the concepts needed for this challenge? Either way, I would recommend tackling more problems to improve your understanding of Ruby mechanics. The more you code, the better you'll perform under time constraints. Keep coding and keep learning!

PR #13, opened 2021-10-20
Eric Chon

echon006:


lib/car.rb
```diff
@@ -0,0 +1,28 @@
...
+              :model
+
+
+
+  def initialize(name, monthly_payment, loan_length)
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @name = ("#{@make}#{@model}")
+    @make = "Ford"     # I couldnt figure out how to make this dynamic so I
+    @model = "Mustang" # hard coded it so i could move forward
```
kat3kasper, line 15: There are a couple options you could have used to make it dynamic:

@make = name.split[0]
@model = name.split[1]

@make = name.split.first
@model = name.split.last

spec/dealership_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    expect(@dealership.cars_by_make("Toyota")).to eq([@car_1])
+  end
+
+  it "checks the total value of the inventory" do
+    @dealership.add_car(@car_1)
+    @dealership.add_car(@car_2)
+    @dealership.add_car(@car_3)
+    @dealership.add_car(@car_4)
+    expect(@dealership.total_value).to eq(156000)
+  end
```
kat3kasper, line 92: I like that via TDD you still set up this test

spec/dealership_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      it "checks the inventory count" do
+        @dealership.add_car(@car_1)
+        @dealership.add_car(@car_2)
+        expect(@dealership.inventory_count).to eq(2)
+      end
+
+end
+
+RSpec.describe Car do
```
kat3kasper, line 46: Since you are in the `dealership_spec` I would stick to only describing the Dealership rather than the Car here

spec/car_spec.rb
```diff
@@ -37,4 +37,15 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it "shows color as nil" do
+    car = Car.new("Ford Mustang", 1500, 36)
+    expect(car.color).to eq(nil)
+  end
+
```
kat3kasper, line 45: Like that you included a nil check for the color!

lib/dealership.rb
```diff
@@ -0,0 +1,41 @@
...
+    def has_inventory?
+      if inventory_count == 1 || inventory_count > 1
+        true
+      else
+        false
+      end
+    end
+
+    def cars_by_make(car_make) #due to hardcoded this wont work.
+      @inventory.find_all do |correct_make|
```
kat3kasper, line 29: So when you are iterating through the inventory you are accessing a car which might help with some of the confusion:

```
@inventory.find_all do |car|
....
```

so now that you know you are going through the cars you can call the `make` method on them. (I realize it won't work as expected with the hard coding)

```
@inventory.find_all do |car|
  car.make == car_make
end
```

this allows us to compare the make of the cars we are iterating through with the make `car_make` that we are passing through to this method

lib/dealership.rb
```diff
@@ -0,0 +1,41 @@
...
+        correct_make.car.make
+      end.size
+    end
+
+    def total_value
+      total = 0 # no time need to first set up something to return car.total_length
+      @inventory.each do |add_car_value|
```
kat3kasper, line 36: This is similar to the above. When you are iterating through @inventory you are accessing each `Car` in the array.

```
@inventory.each do |car|
   total += car.total_cost
end
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    expect(@dealership.inventory).to eq([])
+  end
+  it "checks the how many objects are in the inventory" do
+    expect(@dealership.inventory_count).to eq(0)
+  end
+
+    before :each do
+      @car_2 = Car.new("Ford Mustang", 1000, 48)
+      @car_1 = Car.new("Toyota Prius", 1500, 36)
+    end
```
kat3kasper, line 26: Really like the before each setup you have for the tests!

kat3kasper (commented):
Good effort Eric. I would have liked to see multiple commits while you were working rather than 1 large commit at the end so I could see your process a bit better. I'd spend a bit of extra time working with Enumerators to have a better sense of how they work.

PR #14, opened 2021-10-20
ruben pasillas

shasuna:


lib/car.rb
```diff
@@ -0,0 +1,40 @@
...
+    @color = nil
+  end
+
+  def make # Return @make value
+    @make
+  end
+
+  def model # Return @model value
+    @model
+  end
```
rtillies, line 18: This comment also goes for the `color` method on lines 32-34:
Can you add something near the beginning of this file so that you would not need explicit methods to return these attributes? Hint: it is present in your `dealership.rb` class.

lib/dealership.rb
```diff
@@ -0,0 +1,69 @@
...
+    @address = address
+    @inventory = []
+    @inventory_count = 0
+    @dealership_value = 0
+  end
+
+  def add_car(car)
+    @inventory << car # Shift argument into @inventory array
+    @inventory_count = @inventory.length # Refresh value of @inventory_count
+  end
```
rtillies, line 16: Can you think of a different way to return the `inventory_count` besides updating it every time you add a car? While I know it's not part of this interaction pattern, but what if you had a `sell_car` method that removes a car from your inventory? A potential answer is to also update the `inventory_count` when you sell a car; however, you now have two different methods that explicitly update the `inventory_count`. We generally want to avoid writing duplicate code.

lib/dealership.rb
```diff
@@ -0,0 +1,69 @@
...
+      "address" => @address
+
+    }
+  end
+
+  def average_price_of_car
+
+    self.total_value / @inventory.count # Divide total value by total amount of cars
+      # Note: Not sure how to get this to return as string with a comma in the appropriate location (yet)
+  end
```
rtillies, line 58: Great start with finding the average price of the car as a number. Think about the pattern of adding commas to numbers. Where would you place a comma for a 4-digit number? 5-digit number? 6-digit, 7-digit, 8-digit?

spec/car_spec.rb
```diff
@@ -37,4 +37,18 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
```
rtillies, line 40: Nice tests for colors!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,135 @@
+require 'rspec'
+require './lib/car'
+require './lib/dealership'
+
+RSpec.describe Dealership do
+
+  it 'is an instance of Dealership' do
+      dealership = Dealership.new("Acme Auto", "123 Main Street")
+
```
rtillies, line 9: What's missing in this test?

rtillies, line 9: What's missing in this test?

spec/dealership_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+      car_3 = Car.new("Toyota Tercel", 500, 48)
+      car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+
+      dealership.add_car(car_1)
+      dealership.add_car(car_2)
+      dealership.add_car(car_3)
+      dealership.add_car(car_4)
+
+      expect(dealership.average_price_of_car).to eq(39000)
+    end
```
rtillies, line 117: As discussed in my comment in your `dealership.rb` class, this works as a number, but not yet as a string. Come back to refactor this test as a string (per the interaction pattern), then revisit the code to get it to work properly.

rtillies (commented):
Overall, this looks good. I feel that you would have completed this challenge successfully given a little more time. Keep coding and keep learning!

rtillies (commented):
Overall, this looks good. I feel that you would have completed this challenge successfully given a little more time. Keep coding and keep learning!

PR #15, opened 2021-10-20
Brad Breiten Car Dealership Independent Challenge

jbreit88:
Completed iteration 4

### DIFF
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -12,4 +12,8 @@ class Dealership
     @inventory.length
   end
 
+  def add_car(car)
+    @inventory << car
+  end
+
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -45,4 +45,34 @@ class Dealership
     }
   end
 
+  def average_price_of_car
+    average_cost = total_value / inventory_count
+    format_number(average_cost)
+  end
+
+  def format_number(number)
+    group_nums = number.to_s.chars.to_a.reverse.each_slice(3) #
+    group_nums.map(&:join).join(',').reverse
+  end
+
+  def cars_sorted_by_price
+    sorted = @inventory.sort_by {|car| car.total_cost}
+  end
+
+  def inventory_hash # So ugly, iterates through @inventory like 3 times, can clean up, but out of time.
+    inventory_hash = Hash.new
+    car_makes = []
+
+    @inventory.each do |car|
+      car_makes << car.make
+    end
+
+    no_duplicates = car_makes.uniq
+
+    no_duplicates.each do |make|
+      inventory_hash[make] = cars_by_make(make).map{|car| car}
+    end
+
+    inventory_hash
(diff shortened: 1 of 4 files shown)


lib/dealership.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def cars_by_make(make)
+    cars_by_make_catcher = []
+    @inventory.each do |car|
+      if car.make == make
+        cars_by_make_catcher << car
+      end
+    end
+    cars_by_make_catcher
+  end
```
kat3kasper, line 31: Yes, totally! Another way you could also do this is with the `find_all` enumerator. 

```
def cars_by_make(make)
  @inventory.find_all do |car|
    car.make == make
  end
end
```

lib/dealership.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def average_price_of_car
+    average_cost = total_value / inventory_count
+    format_number(average_cost)
+  end
+
+  def format_number(number)
+    group_nums = number.to_s.chars.to_a.reverse.each_slice(3) #
+    group_nums.map(&:join).join(',').reverse
+  end
```
kat3kasper, line 56: This works! It can be nice to comment where you found this to give credit to the original source and point others to the article if they want to read more about it. Sometimes that is nice when you come across something a little tricky in source code and to credit the source.

lib/dealership.rb
```diff
@@ -0,0 +1,86 @@
...
+    sorted = @inventory.sort_by {|car| car.total_cost}
+  end
+
+  def inventory_hash
+    # Refactored just in time!!
+    inventory_hash = Hash.new
+    @inventory.each do |car|
+      inventory_hash[car.make] = cars_by_make(car.make).map{|car| car}
+    end
+    inventory_hash
```
kat3kasper, line 68: I don't think you have learned about Big O(n) or optimization yet, but just to push you a bit in your learning -- this is a lot of iterations over our inventory. Basically for each car, you are looping through all the cars again. Another way to do this that is more efficient but a little more verbose could be:

```
def inventory_hash
  inventory_hash = Hash.new
  inventory.each do |car|
    if inventory_hash[car.make]
      inventory_hash[car.make] = inventory_hash[car.make] << car
    else
    end
  end
end
```

or you can use the group_by enumerator

```
def inventory_hash
  @inventory.group_by { |car| car.make }
end
```

spec/car_spec.rb
```diff
@@ -2,39 +2,40 @@
 require './lib/car'
 
 RSpec.describe Car do
+  before(:each) do
+    @car = Car.new("Ford Mustang", 1500, 36)
+  end
```
kat3kasper, line 7: Love the use of `before(:each)` here!

kat3kasper (commented):
Awesome work on this Brad! 🎉  Appreciate the small commits as well as you worked on this

PR #16, opened 2021-10-20
IC 2 

Henchworm:
Looking forward to the solution for final method of iteration 4-- was so close!

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -5,10 +5,10 @@ class Car
               :make,
               :color
   def initialize(model, monthly_payment, loan_length)
-    @model           = model[5..11]
+    @model           = model.split(/\W+/).last
     @monthly_payment = monthly_payment
     @loan_length     = loan_length
-    @make            = model[0..3]
+    @make            = model.split(/\W+/).first
     @color           = nil
   end
 
@@ -18,6 +18,6 @@ class Car
 
   def paint!(color)
     @color = color
-  end 
+  end
 
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,23 @@
+class Car
+  attr_reader :model,
+              :monthly_payment,
+              :loan_length,
+              :make,
+              :color
+  def initialize(model, monthly_payment, loan_length)
+    @model           = model[5..11]
+    @monthly_payment = monthly_payment
+    @loan_length     = loan_length
+    @make            = model[0..3]
+    @color           = nil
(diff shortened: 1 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,23 @@
+class Car
+  attr_reader :model,
+              :monthly_payment,
+              :loan_length,
+              :make,
+              :color
+  def initialize(model, monthly_payment, loan_length)
+    @model           = model.split(/\W+/).last
```
kat3kasper, line 8: I like how you set @model and @make right in the initialize!

lib/dealership.rb
```diff
@@ -0,0 +1,69 @@
...
+    @inventory << car
+    @inventory_count += 1
+  end
+
+  def has_inventory?
+    if @inventory.count > 0
+      true
+    else
+      false
+    end
```
kat3kasper, line 23: you could replace this with 

```
def has_inventory?
  @inventory.count > 0
end
```

And if you are going to use the `.count` method anyway, it probably would be easier to set up a method `inventory_count` rather than using an attribute:

```
def inventory_count
   @inventory.count
end
```

spec/car_spec.rb
```diff
@@ -37,4 +37,12 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it "has a color" do
+    car = Car.new("Ford Mustang", 1500, 36)
+    expect(car.color).to eq(nil)
```
kat3kasper, line 43: I like that you checked for nil before painting the car!

lib/dealership.rb
```diff
@@ -0,0 +1,69 @@
...
+    @inventory.sort_by do |car|
+      car.monthly_payment * car.loan_length
+    end
+  end
+
+  def inventory_hash
+    inventory_hash = {}
+    car_array = []
+    @inventory.map do |car|
+      inventory_hash[car.make] = car_array.uniq << car
```
kat3kasper, line 60: Since you are manipulating the inventory_hash within your iteration rather than returning the result from map, I might have used .each instead of map. 

```
    inventory_hash = {}
    @inventory.each do |car|
      if inventory_hash[car.make]
        inventory_hash[car.make] = inventory_hash[car.make] << car
      else
      end
    end
    inventory_hash
```

kat3kasper (commented):
Great job Christopher! Only missed that last `inventory_hash` piece

PR #17, opened 2021-10-20
Josh Walsh

jaw772:


lib/car.rb
```diff
@@ -0,0 +1,22 @@
+class Car
+  attr_accessor :car, :monthly_payment, :loan_length
+  attr_accessor :color
```
kat3kasper, line 3: For now, we should probably keep these as `attr_reader` only as we may not want to expose a write method on the Car class.

lib/car.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize(car, monthly_payment, loan_length)
+    @car = car.split(" ")
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = color
```
kat3kasper, line 8: I would probably set this to nil rather than color since color is undefined.

lib/dealership.rb
```diff
@@ -0,0 +1,58 @@
+class Dealership
+  attr_reader :dealer, :address
+  attr_accessor :inventory, :dealer_info, :inv_hash
```
kat3kasper, line 3: We probably don't want access to any of these as getters/setters and rather as methods we define on the Dealership class.

lib/dealership.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+  def cars_sorted_by_price
+    @inventory.sort_by{|car| car.total_cost}
+  end
+  def inventory_hash
+    @inv_hash["Ford"] = cars_by_make("Ford")
+    @inv_hash["Toyota"] = cars_by_make("Toyota")
+    @inv_hash["Chevrolet"] = cars_by_make("Chevrolet")
+    @inv_hash
+  end
```
kat3kasper, line 57: Would have loved to see some dynamic programming here rather than hardcoded makes.

lib/dealership.rb
```diff
@@ -0,0 +1,58 @@
...
+    @inventory.each do |car|
+      @inv_cost += car.total_cost
+    end
+    @inv_cost
+  end
+  def details
+    @dealer_info["total_value"] = total_value
+    @dealer_info["address"] = address
+    @dealer_info
+  end
```
kat3kasper, line 44: For both `details` and `inventory_hash` it makes a bit more sense if both of those methods return what we are looking for rather than storing the results in instance variables `@dealer_info` and `@inv_hash`.  How it is currently written if you wrote the following code it would not work as expected.

```
> dealership = Dealership.new("Acme Auto", "123 Main Street")
> car_1 = Car.new("Ford Mustang", 1500, 36)
> car_2 = Car.new("Toyota Prius", 1000, 48)
> dealership.add_car(car_1)
> dealership.add_car(car_2)

> dealership.dealer_info 
> {} # Where we actually want { "total_value" => ..., "address" => "..." }

```
It would return the empty hash rather than what you might expect. This is because `@dealer_info` only gets set after calling the `.details` method. 

So you would want something closer to:

```
class Dealership
  attr_reader :dealer, :address, :inventory
  def initialize(dealer, address)
    @dealer = dealer
    @address = address
    @inventory = []
  end

  def details
    dealer_info = {}
    dealer_info["total_value"] = total_value
    dealer_info["address"] = @address
    dealer_info
  end
end
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+  it "checks if there is inventory" do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+
+    expect(dealership.has_inventory?).to be false
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    expect(dealership.has_inventory?).to be true
```
kat3kasper, line 50: Nice check on the before and after!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    expect(dealership.inventory_hash).to eq dealership.inv_hash
```
kat3kasper, line 134: Rather than checking the `dealership.inv_hash` == `dealership.inventory_hash`, we probably want to hard code in the tests what we expect `dealership.inventory_hash` to equal.

```
expect(dealership.inventory_hash.to eq( {"Ford": [car1], "Toyota": [car2, car3], "Chevrolet": [car4]})
```

lib/dealership.rb
```diff
@@ -0,0 +1,58 @@
...
+  def average_price_of_car
+    @average = total_value / inventory_count
+    @average.to_s.insert(2, ",")
+  end
+  def cars_sorted_by_price
+    @inventory.sort_by{|car| car.total_cost}
```
kat3kasper, line 50: Good use of `sort_by`!

lib/dealership.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+    end
+    @makes
+  end
+  def total_value
+    @inv_cost = 0
+    @inventory.each do |car|
+      @inv_cost += car.total_cost
+    end
+    @inv_cost
```
kat3kasper, line 38: `inv_cost` doesn't need to be an instance variable here and can be a local variable:

```
def total_value
    inv_cost = 0
    @inventory.each do |car|
      inv_cost += car.total_cost
    end
    inv_cost
```

Instance variables, those that start with a @ are more helpful in a class when they are being shared across multiple methods within the class. If you only need them for one specific case you should use a local variable.

kat3kasper (commented):
This is a good start Josh. I would take a look back at instance vs local variables, when to use `attr_reader` vs `attr_writer`, and when to use `attr_reader` vs methods on a class. These will really help you in the future when defining objects via Classes.

PR #18, opened 2021-10-20
Eldridge Turambi IC #2

Eldridge-Turambi:
#average_price_of_car method was hard for me because of the string form and added comma to the number 

#inventory_hash method: Is there an easier way to do this??? Got it to work though. 

Had to make helper method #cars_in_category(category) to make my #inventory_hash method cleaner. Hope this helps

### DIFF
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -36,4 +36,43 @@ class Dealership
     }
     details_hash
   end
+
+  def average_price_of_car
+    average_price = (total_value / @inventory.length).to_s
+    first_two = average_price.slice(0..1)
+    last_two = average_price.slice(2..4)
+
+    "#{first_two},#{last_two}"
+
+    ## I FOUND THIS BELOW METHOD ONLINE THAT IS REALLY DYNAMIC. I DO NOT UNDERSTAND IT BUT WOULD LIKE TO LEAVE IT HERE SO I CAN TRY TO UNDERSTAND LATER. THIS IS A HARD METHOD. PLEASE HELP ME UNDERSTAND THIS SOMETIME
+
+    # number_string_format = (total_value / @inventory.length).to_s.chars.reverse.each_slice(3)
+    # number_string_format.map(&:join).join(',').reverse
+  end
+
+  def cars_sorted_by_price
+    @inventory.sort_by do |car|
+      car.total_cost
+    end
+  end
+
+  def cars_in_categoty(category)
+    @inventory.find_all do |car|
+      car.make == category
+    end
+  end
+
+  def inventory_hash
+    hash_of_cars = {}
+    # keys are make name of cars in string form
+    keys = @inventory.map do |car|
+      car.make
+    end
+
+    keys.each do |key|
+      hash_of_cars[key] = cars_in_categoty(key)
+    end
+    hash_of_cars
+  end
+
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -16,4 +16,24 @@ class Dealership
   def inventory_count
     @inventory.length
   end
+
+  def cars_by_make(make_of_car)
+    @inventory.find_all do |car|
+      car.make == make_of_car
+    end
+  end
+
+  def total_value
+    @inventory.sum do |car|
+      car.total_cost
+    end
+  end
+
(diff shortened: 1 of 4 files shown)


spec/car_spec.rb
```diff
@@ -37,4 +36,17 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has nil color by default' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to eq(nil)
+  end
```
kat3kasper, line 44: I like the nil test!

lib/dealership.rb
```diff
@@ -0,0 +1,78 @@
...
+    average_price = (total_value / @inventory.length).to_s
+    first_two = average_price.slice(0..1)
+    last_two = average_price.slice(2..4)
+
+    "#{first_two},#{last_two}"
+
+    ## I FOUND THIS BELOW METHOD ONLINE THAT IS REALLY DYNAMIC. I DO NOT UNDERSTAND IT BUT WOULD LIKE TO LEAVE IT HERE SO I CAN TRY TO UNDERSTAND LATER. THIS IS A HARD METHOD. PLEASE HELP ME UNDERSTAND THIS SOMETIME
+
+    # number_string_format = (total_value / @inventory.length).to_s.chars.reverse.each_slice(3)
+    # number_string_format.map(&:join).join(',').reverse
```
kat3kasper, line 50: I can explain it a bit breaking down by each step:

(total_value/@inventory.length) = average price for ex. 42000

42000.to_s -> "42000"
"42000".char -> ['4', '2', '0', '0', '0']

Now we reverse the array because we know we need a comma every 3 digits when writing American English style numbers. Essentially we want to count from the back of the above array. The easiest way to do that is to first reverse the array:

['4', '2', '0', '0', '0'].reverse = ['0', '0', '0', '2', '4']

now we can group by the 3 digits we need

['0', '0', '0', '2', '4'].each_slice(3) = creates an enumerator when the value your iterating through is an array of up to 3 items [ [0,0,0], [2,4] ] 

Now we wish to put commas between those groupings of digits:

[ [0,0,0], [2,4] ] .map(&:join) = ["000", "24"]

 ["000", "24"].join(',') = "000,24"
 
then reversing that entire string will give us "42,000"

Definitely an advanced topic and good on you for putting in comments that you found it but wasn't really sure why it worked the way it did.

lib/dealership.rb
```diff
@@ -0,0 +1,78 @@
...
+    hash_of_cars = {}
+    # keys are make name of cars in string form
+    keys = @inventory.map do |car|
+      car.make
+    end
+
+    keys.each do |key|
+      hash_of_cars[key] = cars_in_categoty(key)
+    end
+    hash_of_cars
```
kat3kasper, line 75: This definitely works! Another way you could do it without creating a `cars_in_categoty` method is the following:

```
  def inventory_hash
    hash_of_cars = {}
    @inventory.each do |car|
       if hash_of_cars[car.make]
          hash_of_cars[car.make] << car
       else
          hash_of_cars[car.make] = [car]
       end
    end
    hash_of_cars
  end
```
  

Lots of way to write different code to achieve the same outcome :)

lib/dealership.rb
```diff
@@ -0,0 +1,78 @@
...
+    "#{first_two},#{last_two}"
+
+    ## I FOUND THIS BELOW METHOD ONLINE THAT IS REALLY DYNAMIC. I DO NOT UNDERSTAND IT BUT WOULD LIKE TO LEAVE IT HERE SO I CAN TRY TO UNDERSTAND LATER. THIS IS A HARD METHOD. PLEASE HELP ME UNDERSTAND THIS SOMETIME
+
+    # number_string_format = (total_value / @inventory.length).to_s.chars.reverse.each_slice(3)
+    # number_string_format.map(&:join).join(',').reverse
+  end
+
+  def cars_sorted_by_price
+    @inventory.sort_by do |car|
```
kat3kasper, line 54: Great use of `sort_by`!

kat3kasper (commented):
Awesome work with getting through all 4 Iterations Eldridge!! 🎉

PR #20, opened 2021-10-20
Robin Dittrich

Dittrir:


lib/car.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  end
+
+  def make
+    make_and_model.split.first
+  end
+
+  def model
+    make_and_model.split.last
+  end
```
BrianZanti, line 17: This works! Great job using the split method. One thing to consider here is what if you needed to change the make? Or the model? It's difficult because these two attributes are tied together in `make_and_model`. Something you could consider doing instead is splitting them in the initialize method so that they are always separate. Not necessary for this challenge but something to think about.

lib/car.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def color
+    @color
+  end
+
+  def paint!(paint_color)
+    @color = paint_color
+    p @color
+  end
```
BrianZanti, line 30: Looks great!

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def add_car(car)
+    @inventory << car.make + " " + car.model # I know this is a SUPER round about way, but the @make_and_model wasn't working.
+    @number_in_inventory += 1
+    @total_value += car.total_cost
+  end
+
+  def inventory_count
+    p @number_in_inventory
+  end
```
BrianZanti, line 20: I see a couple of places in your code where you have a `p` at the end of your methods. `p` will both print AND return the value. We don't need our code to print the value, we just need it to return the value so you can take the `p` out and just reference the value you want to return. In this case, the method body could just be `@number_in_inventory`

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
+class Dealership
+  attr_reader :name, :location, :inventory, :total_value
+  def initialize(name, location)
+    @name = name
+    @location = location
+    @inventory = []
+    @number_in_inventory = 0
```
BrianZanti, line 7: It's good practice to not create variables if you don't need to. In this case, you don't **need** the `@number_in_inventory` instance variable. Instead, whenever you needed that data you could call `@inventory.count` or `@inventory.length`. The less variables you have to keep track of, the better.

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+    @total_value = 0
+  
+  end
+
+  def add_car(car)
+    @inventory << car.make + " " + car.model # I know this is a SUPER round about way, but the @make_and_model wasn't working.
```
BrianZanti, line 13: If you look at the interaction pattern, we want the Dealership's inventory to contain Car objects. This is different than the String of the car's make and model. If you only have time to review one thing from this challenge I would make this it. This is a very important concept.

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+    else
+      false
+    end
+  end
+
+  def cars_by_make(make) #Dont know why this doesnt work...
+    @inventory.select do |car|
+      car.make == make
+    end
+  end
```
BrianZanti, line 34: This doesn't work because `@inventory` doesn't contain Car objects (see my other comment). Instead it contains Strings and you can't call `.make` on a String, but you can call it on a Car.

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def total_value
+    p @total_value
+  end
+
+  def details
+    details = {'total_value' => @total_value, 'address' => @location}
```
BrianZanti, line 41: Looks good!

spec/car_spec.rb
```diff
@@ -37,4 +37,18 @@
...
+    expect(car.color).to eq(nil)
+  end
+
+  it 'paints the car a color' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    car.paint!(:blue)
+
+    expect(car.color).to eq(:blue)
+  end
```
BrianZanti, line 53: These tests look great 👍

spec/dealership_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  it 'adds cars to inventory' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+
+    expect(dealership.inventory).to eq(["Ford Mustang", "Toyota Prius"])
```
BrianZanti, line 31: This is related to my other comment about adding Car objects. We don't want this method to return Strings, we want it to return Car objects. 

```ruby
expect(dealership.inventory).to eq([car_1, car_2])
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  #Iteration 3
+
+  it 'checks if specific dealership has inventory' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+
+    expect(dealership.has_inventory?).to eq(false)
+  end
```
BrianZanti, line 53: It'd also be a good idea to test when this method returns true.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.cars_by_make("Toyota")).to eq(["Toyota Prius", "Toyota Tercel"])
+    expect(dealership.cars_by_make("Ford")).to eq(["Ford Mustang"])
```
BrianZanti, line 68: Same thing here. We want this method to return Car objects.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.details).to eq({"total_value" => 156000, "address" => "123 Main Street"})
+  end
+end
```
BrianZanti, line 100: Overall your testing looks good!

BrianZanti (commented):
Hey Robin, I left some comments for you to review. I think the most important thing I want you to look at is this idea of the Dealership keeping Car objects in its inventory array and how this is different than an array of Strings. Overall, I think you did a great job on this challenge! 🎉

PR #21, opened 2021-10-20
Wade Naughton

WadeNaughton:


WadeNaughton:
Wade Naughton

lib/car.rb
```diff
@@ -0,0 +1,29 @@
...
+  def model
+    @make.split[1]
+  end
+
+  def total_cost
+    @loan_length * @monthly_payment
+  end
+
+  def paint!(color)
+    @color = []
```
rtillies, line 25: This method should keep track of a single color, not necessarily collect each color in an array

lib/dealership.rb
```diff
@@ -0,0 +1,33 @@
+class Dealership
+
+  attr_reader :name, :address, :inventory
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+  end
```
rtillies, line 9: Your initialize method for dealership looks good!

lib/dealership.rb
```diff
@@ -0,0 +1,33 @@
...
+   @inventory.count
+  end
+
+  def add_car(car)
+    @inventory << car
+  end
+
+  def has_inventory?
+    @inventory == 0
+  end
```
rtillies, line 21: because @inventory is an array, it looks like you want the .count of this array. Instead, you are checking for the value of @inventory, which will never be 0.

Do you already have a method that would check the number of cars in inventory?

lib/dealership.rb
```diff
@@ -0,0 +1,33 @@
...
+  # def cars_by_make(make)
+  #   cars_in_make = []
+  #   @make.each do |make|
+  #     if make.car == car
+  #       cars_in_make << make
+  #     end
+  #
+  #   end
+  #   cars_in_make
+end
```
rtillies, line 33: Sounds like you ran out of time, but looks like you're on the right track!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rspec'
+require './lib/dealership'
```
rtillies, line 2: When I run dealership_spec.rb by itself, I get an uninitialized Constant error for the Car object. Because your dealership uses Car objects, can you figure out which require statement is missing?

spec/dealership_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  # it 'can add car' do
+  #   car = Car.new("Ford Mustang", 1500, 36)
+  #   car_2 = Car.new("Toyota Prius", 1000, 48)
+  #   dealership = Dealership.new("Acme Auto", "123 Main Street")
+  #   dealership.add_car(car, car_2)
+  #
+  #   expect(dealership.inventory).to eq(car, car_2)
+  # end
+end
```
rtillies, line 31: Sounds like you ran out of time, but looks like you're on the right track! A few minor tweaks, but I'm confident you would have solved this test!

rtillies (commented):
You started strong and it looks like you would have progressed further with additional time. I would recommend reviewing this project with a mentor. Overall, keep coding and keep practicing!

PR #22, opened 2021-10-20
Parker Lockhart

ParkerLockhart:


lib/car.rb
```diff
@@ -0,0 +1,20 @@
+class Car
+  attr_reader :makemodel, :make, :model, :monthly_payment, :loan_length, :color
```
dcoleman21, line 2: You wouldn't need to add `makemodel` to your `attr_reader` because we only really ever call `make` or `model` separately.

lib/car.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def initialize(makemodel, monthly_payment, loan_length)
+    @makemodel = makemodel.split
+    @make = @makemodel[0]
```
dcoleman21, line 6: Nice job recognizing that you would need to split that string.

lib/car.rb
```diff
@@ -0,0 +1,20 @@
...
+    @model = @makemodel[1]
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = paint!(:color)
```
dcoleman21, line 10: Does this give you an initial `nil` value?

lib/dealership.rb
```diff
@@ -0,0 +1,42 @@
...
+  def add_car(car)
+    @inventory << car
+  end
+
+  def has_inventory?
+    @inventory.size > 0
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |car|
```
dcoleman21, line 23: Nice use of `.select` here.

lib/dealership.rb
```diff
@@ -0,0 +1,42 @@
...
+      car.make == make
+    end
+  end
+
+  def total_value
+    @inventory.map do |car|
```
dcoleman21, line 29: Nice job using `.map` and then tagging on the `.sum` method at the end of the block. Could you also use the `.sum` enumerable vs the `.map` ?

lib/dealership.rb
```diff
@@ -0,0 +1,42 @@
...
+      car.total_cost
+    end.sum
+  end
+
+  def average_price_of_car
+    (total_value / inventory.size).to_s.insert(-4, ",")
```
dcoleman21, line 35: Nice job solving this. I think this works for this challenge. I also solved it this way but I got to thinking that it might not work if we had a six figure number.

spec/car_spec.rb
```diff
@@ -37,4 +37,9 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has a color' do
```
dcoleman21, line 41: I think you are missing a test here. And that is why your `@color` in the initialize and your `.paint!` methods are working. The interaction pattern stated that a cars color starts out as `nil`
```
pry(main)> car.color
#=> nil

t "starts with no color" do
    car = Car.new("Ford Mustang", 1500, 36)

    expect(car.color).to eq(nil)
  end
```

spec/car_spec.rb
```diff
@@ -37,4 +37,9 @@
...
+    car = Car.new("Ford Mustang", 1500, 36)
+    expect(car.paint!(:blue)).to eq(car.color)
```
dcoleman21, line 43: You would need to add another test that tests that it can get painted:
```
car.paint!(:blue)

it "can get paint to change the color" do
    car = Car.new("Ford Mustang", 1500, 36)
    car.paint!(:mint_green)
    expect(car.color).to eq(:mint_green)
  end
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,95 @@
+require 'rspec'
+require './lib/car'
+require './lib/dealership'
+
+describe Dealership do
+  it 'exists' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    expect(dealership).to be_a Dealership
+  end
+
```
dcoleman21, line 10: I think you are missing a test here:
```
pry(main)> dealership.inventory
#=> []
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    expect(dealership.inventory_count).to eq(0)
+  end
+
+  it 'adds cars' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    expect(dealership.inventory_count).to eq(2)
```
dcoleman21, line 22: You will need to add another expect here:
```
pry(main)> dealership.inventory
#=> [#<Car:0x00007fa53b9ca0a8...>, #<Car:0x00007fccd2985f48...>]
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    expect(dealership.cars_by_make("Toyota").size).to eq(2)
```
dcoleman21, line 40: I think that this expectation is suppose to equal and array.
```
pry(main)> dealership.cars_by_make("Toyota")
#=> [#<car:0x00007fccd2985f48...>, #<car:0x00007fccd383c2d0...>]

```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    expect(dealership.total_value).to eq(156000)
+  end
+
```
dcoleman21, line 55: I think you are missing a test here
```
pry(main)> dealership.details
#=> {"total_value" => 156000, "address" => "123 Main Street"}
```

dcoleman21 (commented):
Overall nice job working through this IC. There were some tricky things. I left feedback for you to review.

PR #23, opened 2021-10-20
Jacob Jenner

jennerj:


lib/car.rb
```diff
@@ -0,0 +1,36 @@
...
+    @loan_length = loan_length
+  end
+
+  def make
+    @make = "Ford"
+  end
+
+  def model
+    @model = "Mustang"
+  end
```
BrianZanti, line 15: With the way these methods are written, a Car object would always return "Ford" for the make and "Mustang" for the model. Instead, we want this class to be dynamic so that Car objects can have any make/model.

lib/car.rb
```diff
@@ -0,0 +1,36 @@
+class Car
+
+  def initialize(make, monthly_payment, loan_length)
+    @make = make
```
BrianZanti, line 4: the `make` argument in the initialize method contains both the make and the model. In the example from the interaction pattern it was "Ford Mustang", but it could be any make/model like "Toyota Camry", "Subaru Forester", etc. So what you could do is separate out the two words in the String and save the first one to `@make` and the second one to `@model`. The String#split method is helpful for breaking down a String into different parts.

lib/car.rb
```diff
@@ -0,0 +1,36 @@
...
+    @model = "Mustang"
+  end
+
+  def monthly_payment
+    @monthly_payment = 1500
+  end
+
+  def loan_length
+    @loan_length = 36
+  end
```
BrianZanti, line 23: It looks like in a lot of your methods you are using the assignment operator `=`. What this will do is **reassign** the value of `@loan_length`. We don't want to do that because that will permanently change that variable. Instead we just want to return the loan length, so this method could be written as

```
def loan_length
  @loan_length
end
```

lib/car.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def total_cost
+    @loan_length.to_f * @monthly_payment.to_f
+  end
+
+  def color
+
+  end
+
+  def paint!
```
BrianZanti, line 33: In the interaction pattern, when we call the `paint!` method we pass it an argument, so you need to write this method so that it accepts an argument: `def paint!(new_color)`

spec/dealership_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  before(:each) do
+    @dealership = Dealership.new("Acme Auto", "123 Main Street")
+    @car_1 = Car.new("Ford Mustang", 1500, 36)
+    @car_2 = Car.new("Toyota Prius", 1000, 48)
+  end
+
+  it 'exists' do
+    # binding.pry
+    expect(@dealership).to be_a(Dealership)
+  end
```
BrianZanti, line 16: This is a good "it exists" test 👍

lib/car.rb
```diff
@@ -0,0 +1,36 @@
...
+    @monthly_payment = 1500
+  end
+
+  def loan_length
+    @loan_length = 36
+  end
+
+  def total_cost
+    @loan_length.to_f * @monthly_payment.to_f
+  end
```
BrianZanti, line 27: Great work!

lib/car.rb
BrianZanti: Also, once this idea clicks, review previous lessons to see how we can do this same thing with an `attr_reader`

BrianZanti (commented):
Hey Jacob, a left you some comments on this IC. Overall, I think you could benefit from reviewing the lesson on Objects, Instances, Classes and deepening your understanding of attributes, instance variables, and attr_readers. I would also recommend reviewing this with a mentor or Leigh.

PR #24, opened 2021-10-20
Seth Perna

SethPerna:
I thought I knew what I was doing, I was wrong. I know I changed the make and model to be two separate arguments.  I should have known a method to get to this point. I am sorry, I tried my best. Super bummed.  Will be reaching out for help from cohort mates to run through this and learn/remind myself what I have forgotten or missed.


lib/car.rb
```diff
@@ -0,0 +1,19 @@
+class Car
+  attr_accessor :make, :model, :monthly_payment, :loan_length, :color
+  def initialize(make, model, monthly_payment, loan_length)
```
dcoleman21, line 3: I saw your comment. So yes, you are correct that you will need to separate out the `make` and `model`. So as it is passed, `Ford Mustang`, is there a method that will `split` a string?
So then you can either pass something like:
```
def initialize (make_model, monthly_payment, loan_length)
   @make = make_model.split(' ').first
   @model = make_model.split(' ').last
  ......
end
```
OR I did it like this:
```
attr_reader :make,
              :model,
              :monthly_payment,
              :loan_length,
              :color

  def initialize(car_type, monthly_payment, loan_length)
    @car_type        = car_type
    @make            = car_type_split.first
    @model           = car_type_split.last
    @monthly_payment = monthly_payment
    @loan_length     = loan_length
    @color           = nil
  end

  def car_type_split
    @car_type.split(" ")
  end
```

lib/car.rb
```diff
@@ -0,0 +1,19 @@
...
+    @make = make
+    @model = model
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = nil
```
dcoleman21, line 8: Nice job setting this to `nil` to start

lib/car.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def total_cost
+    @monthly_payment * @loan_length
+  end
+
+  def paint!(new_color)
+    @color = new_color
```
dcoleman21, line 16: Nice job re-assigning this

lib/dealership.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def add_car(car)
+    @inventory << car
+    @inventory_count += 1
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |make|
+      if car.make == make
```
dcoleman21, line 17: You don't need to use an `if` here. You can just put 
```
car.make == make
```

lib/dealership.rb
```diff
@@ -0,0 +1,22 @@
+class Dealership
+  attr_accessor :name, :address, :inventory, :inventory_count, :car
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+    @inventory_count = 0
```
dcoleman21, line 7: So I think this is a way that you could do the `inventory_count` like this. Here is another possible solution:
```
def inventory_count
    @inventory.length
  end
```

spec/car_spec.rb
```diff
@@ -3,38 +3,52 @@
 
 RSpec.describe Car do
   it 'exists' do
-    car = Car.new("Ford Mustang", 1500, 36)
+    car = Car.new("Ford", "Mustang", 1500, 36)
```
dcoleman21, line 6: You wouldn't want to change your tests by adding in the `,` between the string `"Ford Mustang"`. That will take place in your class.

spec/car_spec.rb
```diff
@@ -3,38 +3,52 @@
...
 
   it 'has a total cost' do
-    car = Car.new("Ford Mustang", 1500, 36)
+    car = Car.new("Ford", "Mustang", 1500, 36)
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it "returns nil for color" do
+    car = Car.new("Ford", "Mustang", 1500, 36)
```
dcoleman21, line 42: These tests look great except what data type did the interaction pattern give us for the `car.paint!`??
```
pry(main)> car.paint!(:blue)
```
😉

spec/car_spec.rb
```diff
@@ -3,38 +3,52 @@
...
+
+    expect(car.color).to eq(nil)
+  end
+
+  it "paints the car" do
+    car = Car.new("Ford", "Mustang", 1500, 36)
+    car.paint!("blue")
+
+    expect(car.color).to eq("blue")
```
dcoleman21, line 51: So adding on to the above comment. What would our expectation be here?
```
pry(main)> car.color
#=> :blue
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+RSpec.describe do
+  it "exists" do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+
+    expect(dealership).to be_a Dealership
+    expect(dealership.inventory).to eq([])
+    expect(dealership.inventory_count).to eq(0)
+  end
+
```
dcoleman21, line 12: I think you are missing a test here
```
pry(main)> dealership.has_inventory?
#=> false
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    car_2 = Car.new("Toyota", "Prius", 1000, 48)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+
+    expect(dealership.inventory).to eq([car_1, car_2])
+    expect(dealership.inventory_count).to eq(2)
+  end
+
+  it "shows cars by make" do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
```
dcoleman21, line 25: Nice job here.

dcoleman21 (commented):
Overall nice job Seth. There were some tricky parts in this one. You did miss a couple of details with the interaction patterns, so just be mindful of that. I left you some feedback to review.

PR #25, opened 2021-10-20
Sierra Tucker

Sierra-T-9598:


lib/car.rb
```diff
@@ -0,0 +1,19 @@
+class Car
+  attr_accessor :monthly_payment, :loan_length, :make, :make_then_model, :model, :color
+  def initialize(make_then_model, monthly_payment, loan_length)
+    @make_then_model = make_then_model
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
```
dcoleman21, line 6: Nice job splitting the `make` and `model` here.

lib/car.rb
```diff
@@ -0,0 +1,19 @@
+class Car
+  attr_accessor :monthly_payment, :loan_length, :make, :make_then_model, :model, :color
```
dcoleman21, line 2: You wouldn't need to add `make_then_model` to your `attr_reader` because we only ever need access to them separately.

lib/car.rb
```diff
@@ -0,0 +1,19 @@
...
+    @model = make_then_model.split[1]
+    @color = nil
+  end
+
+  def total_cost
+    loan_length * monthly_payment
+  end
+
+  def paint!(color)
+    @color = color
```
dcoleman21, line 17: Nice job re-assigning this value

spec/car_spec.rb
```diff
@@ -37,4 +37,17 @@
...
+  it 'has a color' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to eq(nil)
+  end
+
+  it 'gets a paint job' do
+    car = Car.new("Ford Mustang", 1500, 36)
+    # require 'pry'; binding.pry
+    car.paint!(:blue)
```
dcoleman21, line 50: Nice job paying attention to the details here

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
+class Dealership
+  attr_reader :name, :address, :inventory, :inventory_count
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+    @inventory_count = 0
```
dcoleman21, line 7: Nice job here. Another way that you could do `inventory_count` is to create a method and call
```
@inventory.length
```

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def add_car(car)
+    @inventory << car
+    @inventory_count += 1
+  end
+
+  def has_inventory?
+    if @inventory_count == 0
+      false
```
dcoleman21, line 17: Could you just call 
```
@inventory_count > 0
```
and leave off the `if/else`?

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+    else
+      true
+    end
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |car|
```
dcoleman21, line 24: Nice use of `.select` here

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+      car.make == make
+    end
+  end
+
+  def total_value
+    @total = []
```
dcoleman21, line 30: You would not put the `@` sign before your accumulator. You would just write
```
total = [ ]
```
Nice use of `.each` here. Is there an enumerable that sums? 😉

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+    @inventory.each do |car|
+      @total << car.total_cost
+    end
+    @total.sum
+  end
+
+  def details
+    # require 'pry'; binding.pry
```
dcoleman21, line 38: Super close here. So since we are never needing to access this hash, you don't have to set it to a variable. You also do not need to call `@dealership.total_value` and `@dealership.address`. You can just call put those methods in.
```
{
    "total_value" => total_value,
    "address" => @address
}
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,65 @@
+require 'rspec'
+require './lib/car'
+require './lib/dealership'
+
+RSpec.describe Dealership do
+  before(:each)do
+    @dealership = Dealership.new("Acme Auto", "123 Main Street")
```
dcoleman21, line 7: Nice job setting up this before block

spec/dealership_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  it 'adds inventory' do
+    @dealership.add_car(@car_1)
+    @dealership.add_car(@car_2)
+    expect(@dealership.inventory).to eq([@car_1, @car_2])
+    expect(@dealership.inventory_count).to eq(2)
+  end
+
+  it 'has inventory' do
+    expect(@dealership.has_inventory?).to eq(false)
```
dcoleman21, line 33: Nice happy and sad path testing here

dcoleman21 (commented):
Overall, really nice job getting up to the end of I3. You were super close on the details! I left you some feedback to review.

PR #26, opened 2021-10-20
Jimmy Page IC submission

jpage13:


lib/car.rb
```diff
@@ -0,0 +1,22 @@
+class Car
```
rtillies, line 1: Great start!

rtillies (commented):
Hi Jimmy - I would like to see you would move forward throughout the remaining iterations. Because you only made it through Iteration 1, I would want to know how you manage your  time and your understanding of the concepts necessary to complete this challenge. Ultimately, keep coding and keep learning; the more you code, the more comfortable you will get, especially during a timed challenge.

PR #27, opened 2021-10-20
Katy Harrod 

mcharrod:


lib/car.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def model
+    @make_model.split.last
+  end
+
+  def total_cost
+    @loan_length * @monthly_payment
+  end
+end
```
rtillies, line 21: Great start with the Car object. The only thing missing here is the car's paint color and its ability to be painted.

lib/dealership.rb
```diff
@@ -0,0 +1,77 @@
...
+    @inventory << car
+  end
+
+  def inventory_count
+    @inventory.count
+  end
+
+  def has_inventory?
+    inventory_count > 0
+  end
```
rtillies, line 18: Iteration 2 complete. Great job!

lib/dealership.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  # total value pseudocode edition !
+  # inventory holds an array of all Car objects.
+  # for each item in array call total cost method
+  # translate that into an array of values
+  # add up all array of values
+
+  # this was much harder than I anticipated it would be
+  # .....30 mins later....
```
rtillies, line 34: Completely understand your comments here! Sounds like you had the right idea in terms of what you wanted to do, but the crunch to implement within the time limit is real!

lib/dealership.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def cars_sorted_by_price
+      @cost_array = []
+      @inventory.each do |car|
+      @cost_array << car.total_cost
+      @sorted = @cost_array.sort
+      @correct_order = @sorted.reverse()
+      @correct_order
+    end
+  end
```
rtillies, line 62: Given more time, can you step back and understand what you need to change to make this work?

spec/dealership_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    expect(dealership.cars_sorted_by_price).to eq([car_3, car_4, car_2, car_1])
+  end
```
rtillies, line 102: This test doesn't pass successfully. See my comment on your `cars_sorted_by_price` method in your `dealership.rb` file

rtillies (commented):
Great work here, just a missed opportunities that you touched on yourself in your code comments. You ask all of the right questions in class, so keep that up. Keep coding and keep learning!

PR #28, opened 2021-10-20
Russell Rockwood

russellrockwood:


lib/car.rb
```diff
@@ -0,0 +1,28 @@
+#Loan length is in month, and the total cost is the loan length multiplied by the monthly payment.
+class Car
+  attr_accessor :make_model, :monthly_payment, :loan_length, :color
```
BrianZanti, line 3: We don't need to do any writing to these attributes so best to make them `attr_readers`

lib/car.rb
```diff
@@ -0,0 +1,28 @@
...
+    @color = nil
+  end
+
+  def make
+    @make_model.split(' ')[0]
+  end
+
+  def model
+    @make_model.split(' ')[1]
+  end
```
BrianZanti, line 18: This works! Great job using the split method. One thing to consider here is what if you needed to change the make? Or the model? It's difficult because these two attributes are tied together in `make_and_model`. Something you could consider doing instead is splitting them in the initialize method so that they are always separate. Not necessary for this challenge but something to think about.

lib/car.rb
```diff
@@ -0,0 +1,28 @@
+#Loan length is in month, and the total cost is the loan length multiplied by the monthly payment.
+class Car
+  attr_accessor :make_model, :monthly_payment, :loan_length, :color
+
+  def initialize(make_model,monthly_payment, loan_length)
+    @make_model = make_model
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = nil
```
BrianZanti, line 9: Great job 👍

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
+require './lib/car'
+
+class Dealership
+  attr_accessor :name, :address, :inventory, :count
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+    @count = 0
```
BrianZanti, line 10: It's good practice to not create variables if you don't need to. In this case, you don't **need** the `@count` instance variable. Instead, whenever you needed that data you could call `@inventory.count` or `@inventory.length`. The less variables you have to keep track of, the better.

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def add_car(car)
+    @count += 1
+    inventory << car
+  end
+
+  def has_inventory?
+    !@inventory.empty?
```
BrianZanti, line 19: Well done! This is very clean and readable.

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def cars_by_make(make)
+    @inventory.select {|car| car.make == make}
```
BrianZanti, line 23: 👍

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def total_value
+    @inventory.inject(0) {|agg, car| agg + car.total_cost}
```
BrianZanti, line 27: Great job using `inject`! And great job reusing your `total_cost` method. One thing you could improve is naming the variable `agg` to something more readable like `total` or `total_cost`.

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def details
+    dealership_info = Hash.new
+    dealership_info['address'] = self.address
+    dealership_info['total_value'] = self.total_value
+    dealership_info
+  end
```
BrianZanti, line 35: Looks great!

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def average_price_of_car
+    self.total_value / @count
```
BrianZanti, line 38: You're doing a great job reusing your methods.

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
+require './lib/car'
+
+class Dealership
```
BrianZanti, line 3: Your Dealership class is missing a method for `inventory_count`. You have a `count` method in the attr_reader but this is not the same.

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+    dealership_info
+  end
+
+  def average_price_of_car
+    self.total_value / @count
+  end
+
+  def cars_sorted_by_price
+    @inventory.sort {|car| car.total_cost}
+  end
```
BrianZanti, line 43: Close! This would be a good use case for `sort_by`.

spec/car_spec.rb
```diff
@@ -37,4 +37,12 @@
...
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has a color' do
+    car = Car.new("Ford Mustang", 1500, 36)
+    expect(car.color).to eq(nil)
+    
+    car.paint!(:blue)
+    expect(car.color).to eq(:blue)
+  end
```
BrianZanti, line 47: Looks good. You could consider breaking this into to two tests: it "has no color by default" and it "can be painted".

spec/dealership_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  before(:each) do
+    # @car = Car.new
+    @dealership_1 = Dealership.new("Acme Auto", "123 Main Street")
+    @car_1 = Car.new("Ford Mustang", 1500, 36)
+    @car_2 = Car.new("Toyota Prius", 1000, 48)
+    @car_3 = Car.new("Ford Focus", 2000, 24)
+    @car_4 = Car.new("Ford Pinto", 2000, 1)
+  end
+
+  describe "#initialize" do
```
BrianZanti, line 14: I like that you added this extra descriptor.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  describe '#add_car' do
+    it 'adds cars to inventory' do
+      @dealership_1.add_car(@car_1)
+      expect(@dealership_1.count).to eq(1)
+
+      @dealership_1.add_car(@car_2)
+      expect(@dealership_1.inventory[1].model).to eq('Prius')
+    end
+  end
```
BrianZanti, line 31: Your method for adding the car looks good, but I think the test could use some improvement. When you call add car, you want to add a Car object to the Dealership's inventory. So rather than checking the count of cars or the model of cars, check the exact objects you expect to see:

```
expect(@dealership_1.inventory).to eq([@car_1, @car_2])
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  describe '#empty?' do
+    it 'determines if the dealer has inventory' do
+      expect(@dealership_1.has_inventory?).to be(false)
+      @dealership_1.add_car(@car_1)
+      expect(@dealership_1.has_inventory?).to be(true)
+    end
```
BrianZanti, line 37: Great job checking both true and false paths.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  describe '#cars_by_make' do
+    it 'selects cars by manufacturer' do
+      @dealership_1.add_car(@car_1)
+      @dealership_1.add_car(@car_2)
+      @dealership_1.add_car(@car_3)
+
+      expect(@dealership_1.cars_by_make('Ford').length).to eq(2)
+      expect(@dealership_1.cars_by_make('Toyota').length).to eq(1)
```
BrianZanti, line 47: Same thing here. You can expect the exact car objects in the form of `[@car_1, @car_3, @car_4]` for example.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  describe '#cars_sorted_by_price' do
+    it 'sorts cars by price' do
+      @dealership_1.add_car(@car_1)
+      @dealership_1.add_car(@car_2)
+      @dealership_1.add_car(@car_3)
+      @dealership_1.add_car(@car_4)
+      # expect(@dealership.cars_sorted_by_price[0].model).to eq('Pinto')
+    end
+  end
+end
```
BrianZanti, line 86: Overall, testing looks good! It's very organized with describe/it blocks.

BrianZanti (commented):
Hey Russell, great job on this challenge! There were a couple of comments I left on things to improve but overall I think you did really well 🎉  One thing I would emphasize out of all of my comments is that idea of checking the exact objects you expect in your tests.

PR #29, opened 2021-10-20
David Kassin

dkassin:


lib/dealership.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+  avg_car
+  car_sum = 0
+    avg_car.each do |cost|
+      car_sum += cost
+    end
+  avg = car_sum / avg_car.count
+  # Could not find an easy way that I understood to add commas.
+  ## I know I should be using a helper method to calculate the average
+  ## but I was pressed for times.
```
kat3kasper, line 59: One way you could have done the average is to calculate it based on other information you had:

```
total_value/inventory_count
```

Obviously when pressed for time, it can be hard to see that as an option! Nice job with getting out something

lib/dealership.rb
```diff
@@ -0,0 +1,62 @@
...
+    @inventory_count = 0
+  end
+
+  def add_car(car)
+    @inventory << car
+    @inventory_count += 1
+  end
+
+  def has_inventory?
+    @inventory.count > 0
```
kat3kasper, line 21: I would stick with either using @inventory.count or @inventory_count throughout this class.

spec/car_spec.rb
```diff
@@ -37,4 +37,18 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has color' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to eq(nil)
+  end
```
kat3kasper, line 45: Good use of nil test first! I may have called the test something else since it doesn't have a color when initialized. `has color attribute` or `color is nil on itialization`

spec/dealership_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+     dealership = Dealership.new("Acme Auto", "123 Main Street")
+     car_1 = Car.new("Ford Mustang", 1500, 36)
+     car_2 = Car.new("Toyota Prius", 1000, 48)
+     car_3 = Car.new("Toyota Tercel", 500, 48)
+     car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+     dealership.add_car(car_1)
+     dealership.add_car(car_2)
+     dealership.add_car(car_3)
+     dealership.add_car(car_4)
+     expect(dealership.average_price_of_car).to eq ("39,000")
```
kat3kasper, line 94: Love that you still followed TDD and got the test in for this

kat3kasper (commented):
Nice work David! I really appreciated your small commits and your messages. You should see your approach throughout the challenge. 🎉

PR #31, opened 2021-10-20
Hannah Warren

hannahkwarren:


lib/car.rb
```diff
@@ -0,0 +1,33 @@
...
+    total_cost = loan_length * monthly_payment
+    total_cost
+  end
+
+  # def paint!(input)
+  #   paint = input.to_sym
+  #   color << paint
+  #   color
+  # end
+
```
kat3kasper, line 32: You were close with the color!

```
def paint!(input)
  @color = input
end
```

Then in the initializer you could have set `@color = nil` to start.

lib/dealership.rb
```diff
@@ -0,0 +1,40 @@
...
+    @inventory << car
+  end
+
+  def has_inventory?
+    @inventory.count > 0
+  end
+
+  def cars_by_make(given)
+    short_list = @inventory.select {|car| car.make == given}
+    short_list.count
```
kat3kasper, line 25: I think you would have had this if you returned the `short_list` rather than the `count`!

lib/dealership.rb
```diff
@@ -0,0 +1,40 @@
...
+    values = []
+    @inventory.each do |car|
+      values << car.total_cost
+    end
+    values.sum
+  end
+
+  def details
+    details_hash = {"total_value" => self.total_value, "address" => self.address}
+  end
```
kat3kasper, line 38: Exactly! Like this short syntax :)

kat3kasper (commented):
Nice start Hannah! While you were able to work your way through most of the first 3 iterations, the goal of this challenge was to use TDD. Perhaps you forgot to commit the `spec/dealership_spec.rb` file? If so, it's okay we all make that mistake. If not, I would work on figuring how to TDD and go back to those lessons.

lib/car.rb
hannahkwarren: @kat3kasper Yes! I forgot to add the spec for dealer 🙈 it's on [my repo](https://github.com/hannahkwarren/car_dealership_2110) now, though. Sorry about that!

PR #32, opened 2021-10-20
Joel Grant

joel-grant:


lib/car.rb
```diff
@@ -0,0 +1,31 @@
...
+    @loan_length = loan_length
+    @make_model = make_model
+    @color
+  end
+
+  def make
+    make = []
+    make = @make_model.split(" ")
+    make[0]
+  end
```
kat3kasper, line 16: You don't need to set make=[] here since you are immediately overwriting it on the next line. With your approach this variable naming may make more sense:

```
def make
   make_model_array = @make_model.split(" ") #since at this stage it is still both the make and model
   make_model_array[0]
end
```

lib/dealership.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def cars_by_make(make) # try to refactor without if statement if time.
+    car_match = []
+    @inventory.find_all do |car|
+      if car.make == make
+        car_match.push(car)
+      end
+    end
+  end
```
kat3kasper, line 25: I like that you left yourself comments here! 

Some ways to improve this code would be utilizing the `find_all`. You are super close:

```
def cars_by_make(make)
  @inventory.find_all do |car|
    car.make == make
  end
end
```

Your approach is closer to how you might use `.each` rather than `.find_all` since you aren't using the return from `find_all`.

lib/dealership.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def total_value
+    total = @inventory.map do |car|
+      car.total_cost
+    end
+
+    total = total.inject(:+)  # Instructor: is it possible to reach into the car
```
kat3kasper, line 32: Honestly, I rarely use `.inject` but I can't think of an easy way to do this rather than what you did. Another enumerable you could have used though would have been `.sum` https://apidock.com/ruby/v2_5_5/Enumerable/sum

spec/car_spec.rb
```diff
@@ -37,4 +37,20 @@
...
+    end
+  end
+
+  describe '#paint' do
+    it 'applies the paint color to the color object' do
+      car = Car.new("Ford Mustang", 1500, 36)
+      car.paint!(:blue)
+      expect(car.color).to eq(:blue)
+    end
+  end
```
kat3kasper, line 54: I like the nil and paint testing here

spec/dealership_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+    end
+  end
+
+  describe '#cars_by_make()' do
+    it 'returns a list of cars based on make' do
+      dealership = Dealership.new("Acme Auto", "123 Main Street")
+      car_1 = Car.new("Ford Mustang", 1500, 36)
+      car_2 = Car.new("Toyota Prius", 1000, 48)
+      car_3 = Car.new("Toyota Tercel", 500, 48)
+      car_4 = Car.new("Chevrolet Bronco", 1250, 24)
```
kat3kasper, line 45: Can you think of a way to use before each so that you don't need to redefine the cars in every test?

spec/dealership_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+      dealership.add_car(car_1)
+      dealership.add_car(car_2)
+      dealership.add_car(car_3)
+      dealership.add_car(car_4)
+      expect(dealership.inventory_hash).to eq({"Ford"=> [car_1],"Toyota"=> [car_2, car_3],"Chevrolet"=> [car_4]})
+    end
+  end
```
kat3kasper, line 127: I appreciate you sticking to TDD to write up these tests before getting a chance to implement the code!

kat3kasper (commented):
Great job getting through the first 3 iterations Joel! Love the use of commits to make small iterative changes 🎉

PR #34, opened 2021-10-20
Jessica Grazulis

jgrazulis:


lib/car.rb
```diff
@@ -0,0 +1,26 @@
+class Car
+
+  attr_accessor :make, :model, :monthly_payment, :loan_length, :color
+
+  def initialize(make_and_model, monthly_payment, loan_length)
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    make_and_model = {
```
dcoleman21, line 8: Nice thought here. This was tricky because the make and model was coming in as a single string.
``` 
"Ford Mustang"
```
So we would have wanted to split this into two pieces. One way to do this could be to create a method that does the splitting and then call that in the initialize. https://ruby-doc.org/core-2.4.0/String.html#:~:text=split(pattern%3Dnil,4%22%2C%20%22%22%2C%20%22%22%5D%0A%0A%22%22.split(%27%2C%27%2C%20-1)%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%23%3D%3E%20%5B%5D

lib/car.rb
```diff
@@ -0,0 +1,26 @@
...
+      make: @make,
+      model: @model
+      }
+    end
+
+
+  def color
```
dcoleman21, line 15: Pay close attention to the interaction pattern. The return value for `color` was `nil` not and empty array.
```
pry(main)> car.color
#=> nil
```

lib/car.rb
```diff
@@ -0,0 +1,26 @@
...
+    []
+  end
+
+  def total_cost
+    54000
```
dcoleman21, line 20: As it is written, this is called `hard coding`. What variables do we have that we could use to get that value more dynamically?

lib/car.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def paint!(paint_color)
+    color << paint_color
```
dcoleman21, line 24: Since our `color` started out as `nil`, we would want to re-assign `color` to the `paint_color` that you are passing this method.

lib/dealership.rb
```diff
@@ -0,0 +1,7 @@
+class Dealership
+   attr_reader :dealership_name, :address
+
+   def initialize(dealership_name, address)
```
dcoleman21, line 4: Nice start here

spec/car_spec.rb
```diff
@@ -37,4 +37,10 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
```
dcoleman21, line 40: I think you are missing a test in here. We would need to test for the color:
```
pry(main)> car.color
#=> nil
```

spec/car_spec.rb
```diff
@@ -37,4 +37,10 @@
...
+  it 'can be painted' do
+    car = Car.new("Ford Mustang", 1500, 36)
```
dcoleman21, line 42: This test would need to incorporate the `.paint!` method:
```
pry(main)> car.paint!(:blue)
```
and then you could 
```
expect(car.color).to eq(:blue)
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+RSpec.describe Dealership do
+  it 'exists' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+
+    expect(dealership).to be_a Dealership
+  end
+
+  it 'has attributes' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
```
dcoleman21, line 12: Nice job adding this test for the attributes of a dealership even though it wasn't explicitly stated to do so.

dcoleman21 (commented):
Nice job, Jessica. There were a couple of tricky pieces to this challenge. I left you some feedback to review.

PR #35, opened 2021-10-20
Arnaldo Aparicio

arnaldoaparicio:


lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+    total.reduce(:+)
+  end
+
+  def details
+    hash = {}
+    t_value = total_value
+    dealer_address = @address
+    hash["total_value"] = t_value
+    hash["address"] = dealer_address
+    hash
```
kat3kasper, line 38: another way you could simplify this a bit could be:

```
hash = {}
hash["total_value"] = total_value
hash["address"] = @address
hash
```

rather than creating additional local variables.

An even shorter syntax could be:

```
{ "total_value" => total_value, "address" => @address }
```

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+  def add_car(car)
+    @inventory << car
+  end
+
+  def has_inventory?
+    @inventory.length > 0
+  end
+
+  def cars_by_make(make)
+    @inventory.select {|car_make| car_make.make == make}
```
kat3kasper, line 22: Since you are iterating through the cars via `@inventory` rather than the makes, I might change the variable.

```
@inventory.select {|car| car.make == make}
```

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+    t_value = total_value
+    dealer_address = @address
+    hash["total_value"] = t_value
+    hash["address"] = dealer_address
+    hash
+  end
+
+  def average_price_of_car
+    total_value / @inventory.length
+  end
```
kat3kasper, line 43: We were looking for the string to be formatted with a "," here.

kat3kasper (commented):
Nice work Arnaldo with getting through 3 iterations! I see where you were able to get the tests going for iteration 4 so great job following through with TDD

PR #36, opened 2021-10-22
Croix Kimmel IC car_dealershipo_2110

croixk:
Sorry for the delay!!!

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,26 @@
+class Car
+
+  attr_reader :make_model, :monthly_payment, :loan_length
+
+  def initialize(make_model, monthly_payment, loan_length)
+    @make_model = make_model
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+  end
+
+
+  def make
+    make_model_split = make_model.split()
+    make = make_model_split[0]
+  end
+
+  def model
+    make_model_split = make_model.split()
+    model = make_model_split[1]
+  end
+
+  def total_cost
+    total_cost = @monthly_payment * @loan_length
+  end
+
+end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,11 +1,12 @@
 class Car
 
-  attr_reader :make_model, :monthly_payment, :loan_length
+  attr_reader :make_model, :monthly_payment, :loan_length, :color
 
   def initialize(make_model, monthly_payment, loan_length)
     @make_model = make_model
     @monthly_payment = monthly_payment
     @loan_length = loan_length
+    @color = nil
   end
 
 
@@ -21,6 +22,10 @@ class Car
 
   def total_cost
     total_cost = @monthly_payment * @loan_length
+  end
+
+  def paint!(color)
+    @color = color
   end
 
 end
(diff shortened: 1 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,31 @@
+class Car
+
+  attr_reader :make_model, :monthly_payment, :loan_length, :color
+
+  def initialize(make_model, monthly_payment, loan_length)
+    @make_model = make_model
```
kat3kasper, line 6: Since we don't really need access to `make_model` later, we could initialize @make and @model here:

```
make_model = make_model
@make = make_model.split()[0]
@model = make_model.split()[1]
```

just another option :)

lib/dealership.rb
```diff
@@ -0,0 +1,59 @@
...
+  def inventory_count
+    @inventory.count
+  end
+
+  def has_inventory?
+    if @inventory.count != 0
+      return true
+    else
+      return false
+    end
```
kat3kasper, line 26: One way to simplify this could be to return @inventory.count != 0:

```
def has_inventory?
  @inventory.count != 0 
end
```

spec/car_spec.rb
```diff
@@ -37,4 +37,20 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+
+  it 'color' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to eq(nil)
+  end
```
kat3kasper, line 46: Like the nil test here!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+  end
+
+  it 'has_inventory? - false' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    expect(dealership.has_inventory?).to eq(false)
+  end
+
+  # might not need this one? interaction pattern doesn't specify it, but thought
+    # it would be a good check?
+  it 'has_inventory? - true' do
```
kat3kasper, line 44: I think it's good to check both the true and false paths! I might name them `has_inventory? returns false with no cars` and `has_inventory? returns true when there are cars`

spec/dealership_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+    car_2 = Car.new("Toyota Prius", 1000, 48) #48000
+    car_3 = Car.new("Toyota Tercel", 500, 48) #24000
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24) #30000
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.cars_sorted_by_price).to eq([car_3, car_4, car_2, car_1])
+  end
```
kat3kasper, line 142: I appreciate you following TDD and writing the test for this first as well as putting the totals in comments so easy to see the order the cars should be in.

kat3kasper (commented):
Great job getting through the first 3 iterations Croix! I would like to see a few more commits next time rather than the larger commits.

PR #41, opened 2021-12-08
Lisa Taylor 

lisataylor5472:


lib/car.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def make
+    @make_model[0]
+  end
+
+  def model
+    @make_model[1]
+  end
+
```
cjsim89: Missing the `color` and `paint` methods from the interaction pattern's Iteration 1 (e.g. `car.color`, then `car.paint!(arg)`),

PR #43, opened 2021-12-08
Mark Ertmer

markertmer:


lib/car.rb
```diff
@@ -0,0 +1,18 @@
+class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length, :color
+
+  def initialize(make_model, payment, loan_length)
+    @make = make_model.split(" ")[0]
+    @model = make_model.split(" ")[1]
+    @monthly_payment = payment
+    @loan_length = loan_length
+  end
```
rtillies, line 9: Good work using `split` on your `make_model` argument. Can you think of a different way to implement this to prevent having to call `split` multiple times within `initialize`?

lib/dealership.rb
```diff
@@ -0,0 +1,62 @@
...
+    @inventory.count
+  end
+
+  def add_car(car)
+    @inventory << car
+  end
+
+  def has_inventory?
+    @inventory != []
+  end
```
rtillies, line 19: This works. Can you come up with an alternative way to make this work using another method in this class?

lib/dealership.rb
```diff
@@ -0,0 +1,62 @@
...
+    return by_make
+  end
+
+  def total_value
+    tv = 0
+    inventory.each do |car|
+      tv += car.total_cost
+    end
+    return tv
+  end
```
rtillies, line 35: What is `tv`? It seems obvious to you (and me), but what about someone else who is less familiar with Ruby or coding in general?

lib/dealership.rb
```diff
@@ -0,0 +1,62 @@
...
+    makes = @inventory.map do |car|
+      car.make
+    end
+    makes.uniq!
+    hash = {}
+    makes.each do |make|
+      hash[make] = cars_by_make(make)
+    end
+    return hash
+  end
```
rtillies, line 61: Interesting solution here. Good job making this work!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+
+    cars = [car_1, car_2, car_3, car_4]
+    cars.each do |car|
+      dealership.add_car(car)
+    end
+
+    expect(dealership.average_price_of_car).to be 39000
```
rtillies, line 100: Your code passes this test as written. However, does this match the interaction pattern?
```
pry(main)> dealership.average_price_of_car
#=> "39,000"
```

rtillies (commented):
Congrats on nearly completing Iteration 4! I can tell you are using TDD to write good tests and allowing the tests to write quality code. Most of my "critical" comments are nit-picky, but the goal for you is small incremental changes, even at this early stage of your development career. However, you will want to improve on making **ONE** commit throughout the entire exercise. Otherwise, keep learning and keep coding!

rtillies:

PR #44, opened 2021-12-08
Sam Ivari 

samivari:
image

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,8 +1,14 @@
 class Car
   attr_reader :make, :model, :monthly_payment, :loan_length
 
-  def initialize(make_and_model, _monthly_payment, _loan_length)
+  def initialize(make_and_model, monthly_payment, loan_length)
     @make = make_and_model.split[0]
     @model = make_and_model.split[1]
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+  end
+
+  def total_cost
+    loan_length * monthly_payment
   end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,14 +1,11 @@
 class Car
-  attr_reader :make, :model, :monthly_payment, :loan_length
+  attr_reader :make, :model, :monthly_payment, :loan_length, :total_cost
 
   def initialize(make_and_model, monthly_payment, loan_length)
     @make = make_and_model.split[0]
     @model = make_and_model.split[1]
     @monthly_payment = monthly_payment
     @loan_length = loan_length
-  end
-
-  def total_cost
-    loan_length * monthly_payment
+    @total_cost = loan_length * monthly_payment
   end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,11 +1,16 @@
 class Car
-  attr_reader :make, :model, :monthly_payment, :loan_length, :total_cost
+  attr_reader :make, :model, :monthly_payment, :loan_length, :total_cost, :color
 
-  def initialize(make_and_model, monthly_payment, loan_length)
-    @make = make_and_model.split[0]
-    @model = make_and_model.split[1]
+  def initialize(make_n_model, monthly_payment, loan_length)
+    @make = make_n_model.split[0]
+    @model = make_n_model.split[1]
     @monthly_payment = monthly_payment
     @loan_length = loan_length
     @total_cost = loan_length * monthly_payment
+    @color = color
+  end
+
+  def paint!(new_color)
+    @color = new_color
   end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,8 @@
+class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length
+
+  def initialize(make_and_model, _monthly_payment, _loan_length)
+    @make = make_and_model.split[0]
+    @model = make_and_model.split[1]
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,11 @@
+class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length, :total_cost
+
+  def initialize(make_and_model, monthly_payment, loan_length)
+    @make = make_and_model.split[0]
+    @model = make_and_model.split[1]
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @total_cost = loan_length * monthly_payment
```
cjsim89, line 9: Missing `#color` and `#paint` methods in `Car` from I1

lib/dealership.rb
```diff
@@ -0,0 +1,53 @@
+class Dealership
+  attr_reader :name, :address, :inventory, :car
```
cjsim89: Careful with putting `:car` in your attr_reader.... Is `:car` really an attribute of Dealership?

cjsim89: (Trick question, no it is not because you don't use `@car` anywhere in your code) ;)

lib/dealership.rb
```diff
@@ -0,0 +1,53 @@
...
+  def inventory_count
+    inventory.count
+  end
+
+  def add_car(car)
+    inventory << car
+  end
+
+  def has_inventory?
+    return false if inventory.count == 0
```
cjsim89: This returns false, it does not evaluate the `if inventory.count == 0`. Plus, while it is readable, it is not readable by Ruby in the way you expect. I suggest writing what you wrote here as: 
```
if inventory.count == 0
     return false
else 
     return true
end
```

You only wrote one test to check for a `false` inventory, which passes. If you wrote a test that returned `true` for inventory, you would be forced to rewrite this functionality as it does not ever pass for a `true` test.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    expect(dealership.inventory).to eq([car_1, car_2])
+    expect(dealership.inventory_count).to eq(2)
+  end
+
+  # Iteration 3
+  it 'has an inventory' do
+    dealership = Dealership.new('Acme Auto', '123 Main Street')
+    expect(dealership.has_inventory?).to be false
```
cjsim89, line 34: Your `has_inventory?` method will always return `false` or `nil`, because of the way it's written. See the comment I made in the `Car` class file. 

To really test this functionality fully, make a test for both paths this method could return, both true and false. For false, this test is fine (except for the name, this one should be "it does not have an inventory yet") for true, something like this: 
```
it 'has an inventory' do
     dealership = Dealership.new(...)
     dealership.add_car(car_1)
     expect(dealership.has_inventory?).to be true
end
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    car_1 = Car.new('Ford Mustang', 1500, 36)
+    car_2 = Car.new('Toyota Prius', 1000, 48)
+    car_3 = Car.new('Toyota Tercel', 500, 48)
+    car_4 = Car.new('Chevrolet Bronco', 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    expect(dealership.cars_by_make('Toyota')).to eq([car_2, car_3])
+    expect(dealership.details).to eq({ total_value: 156_000, address: '123 Main Street' })
```
cjsim89: This was very close, but you used symbols for `total_value` and `address`, where the interaction pattern was looking for strings. So, `{"total_value" => 156000, "address" => '123 Main Street' }`. 
It may seem like a small thing, but your tests and corresponding code should always follow the interaction patterns exactly.

cjsim89: e.g. 
```
pry(main)> dealership.details
#=> {"total_value" => 156000, "address" => "123 Main Street"
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    dealership = Dealership.new('Acme Auto', '123 Main Street')
+    car_1 = Car.new('Ford Mustang', 1500, 36)
+    car_2 = Car.new('Toyota Prius', 1000, 48)
+    car_3 = Car.new('Toyota Tercel', 500, 48)
+    car_4 = Car.new('Chevrolet Bronco', 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    expect(dealership.average_price_of_car).to eq(39_000)
```
cjsim89: Interaction pattern was looking for a string here, specifically with a comma: 
```
pry(main)> dealership.average_price_of_car
#=> "39,000"
```

lib/dealership.rb
```diff
@@ -0,0 +1,53 @@
...
+      address: address,
+      total_value: total_value
+    }
+  end
+
+  def average_price_of_car
+    average = inventory.map do |car|
+      car.total_cost
+    end
+    average.sum / inventory.count
```
cjsim89: interaction pattern was looking for a string to come out of this method, so you'd want to do a `.to_s` at the end of your line 45. You may also need to find a way to add a thousands' comma (39,000) where necessary. I4s will always push you a little bit. :)

PR #45, opened 2021-12-08
Greg Flaherty - car_dealership_2111

gregoryjflaherty:
Please see completed IC: car_dealership_2111. All tests passing

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,10 +1,11 @@
 class Car
-  attr_reader :make, :model, :monthly_payment, :loan_length
+  attr_reader :make, :model, :monthly_payment, :loan_length, :total_cost
 
   def initialize(car, monthly_payment, loan_length)
     @make = car.split(" ").first
     @model = car.split(" ").last
     @monthly_payment = monthly_payment
     @loan_length = loan_length
+    @total_cost = @monthly_payment * @loan_length
   end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,5 +1,6 @@
 class Car
-  attr_reader :make, :model, :monthly_payment, :loan_length, :total_cost
+  attr_reader :make, :model, :monthly_payment,
+              :loan_length, :total_cost, :color
 
   def initialize(car, monthly_payment, loan_length)
     @make = car.split(" ").first
@@ -7,5 +8,6 @@ class Car
     @monthly_payment = monthly_payment
     @loan_length = loan_length
     @total_cost = @monthly_payment * @loan_length
+    @color = nil
   end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,7 @@
+class Car
+  def initialize(make, monthly_payment, loan_length)
+               @make = make
+    @monthly_payment = monthly_payment
+        @loan_length = loan_length
+  end
+end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,7 +1,9 @@
 class Car
-  def initialize(make, monthly_payment, loan_length)
-               @make = make
+  attr_reader :make, :model, :monthly_payment, :loan_length
+
+  def initialize(car, monthly_payment, loan_length)
+    @make = car.split(" ").first
     @monthly_payment = monthly_payment
-        @loan_length = loan_length
+    @loan_length = loan_length
   end
 end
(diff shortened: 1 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,17 @@
+class Car
+  attr_reader :make, :model, :monthly_payment,
+              :loan_length, :total_cost, :color
+
+  def initialize(car, monthly_payment, loan_length)
+    @make = car.split(" ").first
+    @model = car.split(" ").last
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @total_cost = @monthly_payment * @loan_length
```
kat3kasper, line 10: The only thing I might do is put this in a method rather than an instance var perhaps because it is a calculation. Not a hard and fast rule here though and this does work

```
def total_cost
  @monthly_payment * @loan_length
end
```

lib/dealership.rb
```diff
@@ -0,0 +1,58 @@
+require './lib/car'
```
kat3kasper, line 1: You shouldn't need the Car class in this file

lib/dealership.rb
```diff
@@ -0,0 +1,58 @@
...
+    @inventory.size
+  end
+
+  def add_car(car)
+    @inventory << car
+  end
+
+  def has_inventory?
+    @inventory.size > 0 ? true : false
+  end
```
kat3kasper, line 21: Since @inventory.size > 0 returns a boolean you could shorten this even further

```suggestion
  def has_inventory?
    @inventory.size > 0
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def cars_by_make(make)
+    @inventory.select { |car| car.make == make }
```
kat3kasper, line 24: Nice use of `.select`!

lib/dealership.rb
```diff
@@ -0,0 +1,58 @@
...
+  def average_price_of_car
+    avg_price = total_value / @inventory.size
+    avg_price.to_s.insert(-4, ",")
+  end
+
+  def cars_sorted_by_price
+    @inventory.sort_by do |car|
+      car.total_cost
+    end
+  end
```
kat3kasper, line 49: Love this! Super clean

lib/dealership.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def inventory_hash
+    inventory_hash = Hash.new { |hash, key| hash[key] = [] }
```
kat3kasper, line 52: Interesting. I've never even seen this syntax before

spec/dealership_spec.rb
```diff
@@ -0,0 +1,142 @@
+require 'rspec'
+require './lib/car'
+require './lib/dealership'
+
+RSpec.describe Dealership do
```
kat3kasper, line 5: You may want to look into `before(:each)` so you don't need to redefine 

```
car_1 = Car.new("Ford Mustang", 1500, 36)
car_2 = Car.new("Toyota Prius", 1000, 48)
```

etc. in every test

spec/dealership_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.inventory_hash).to be_a(Hash)
+    expect(dealership.inventory_hash.keys).to eq(["Ford", "Toyota", "Chevrolet"])
+    expect(dealership.inventory_hash.values.flatten.size).to eq(4)
+    expect(dealership.inventory_hash.values.flatten[0]).to be_a(Car)
+    expect(dealership.inventory_hash.values.flatten[1]).to be_a(Car)
+    expect(dealership.inventory_hash.values.flatten[2]).to be_a(Car)
+    expect(dealership.inventory_hash.values.flatten[3]).to be_a(Car)
```
kat3kasper, line 140: This is a little hard to check if our results are right. I might do:

```
expect(dealership.inventory_hash).to eq({"Ford"=> [car_1], "Toyota"=> [car_2, car_3], "Chevrolet"=> [car_4]})
```

kat3kasper (commented):
Really awesome work on this Greg! 🎉 I made a few minor notes, but otherwise, everything looked great. Keep up the detailed commit history 💪

PR #46, opened 2021-12-08
jake_edelstein - car_dealership_2111

jake-edel:
car_dealership_2111 - iterations 1-4

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,5 +1,5 @@
 class Car
-  attr_reader :make, :monthly_payment, :loan_length
+  attr_reader :make, :model, :monthly_payment, :loan_length
 
   def initialize(make_model, monthly_payment, loan_length)
     make_model = make_model.split(' ')
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -8,4 +8,8 @@ class Car
     @monthly_payment = monthly_payment
     @loan_length = loan_length
   end
+
+  def total_cost
+    monthly_payment * loan_length
+  end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,11 @@
+class Car
+  attr_reader :make, :monthly_payment, :loan_length
+
+  def initialize(make_model, monthly_payment, loan_length)
+    make_model = make_model.split(' ')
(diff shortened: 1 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,23 @@
...
+    @loan_length = loan_length
+  end
+
+  def total_cost
+    monthly_payment * loan_length
+  end
+
+  def color
+    @color
+  end
```
kat3kasper, line 18: This is the code we get when we use `attr_reader` so you could have also added color there and gotten the same results without having to write this method. This definitely works though!

lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+    @inventory << car
+  end
+
+  def cars_by_make(make)
+    @inventory.select { |car| car.make == make }
+  end
+
+  def total_value
+    @inventory.sum { |car| car.total_cost }
+  end
```
kat3kasper, line 28: 🙌 Great use of enumerables!!!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,72 @@
+require './lib/car'
+require './lib/dealership'
+
+RSpec.describe Dealership do
+  before(:each) do
```
kat3kasper, line 5: Like the use of `before(:each)`!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      expect(@dealership.details).to eq({"total_value" => 156000, "address" => "123 Main Street"})
+      expect(@dealership.details).to be_instance_of Hash
+    end
+
+    it 'can return the average price of a car' do
+      expect(@dealership.average_price_of_car).to be_instance_of String
+      expect(@dealership.average_price_of_car).to eq "39,000"
+    end
+
+    it 'can return a hash of cars in inventory by make' do
```
kat3kasper, line 65: I would have liked to see the testing for `cars_sorted_by_price` event if you weren't able to get it passing

kat3kasper (commented):
Really great work on this Jake! 🎉  Love the commit history. Keep that up! A few minor comments, but nothing major

PR #47, opened 2021-12-08
Jesse Ochs Car Dealership IC

JesseOchs:
Dang I was so close to finishing the 4th iteration. I have 15 minutes left and wanted to hand this in before it was late.

### DIFF
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -14,4 +14,8 @@ class Dealership
   def add_car(car)
     @inventory << car
   end
+
+  def has_inventory?
+    @inventory.count > 1
+  end
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,7 +1,9 @@
 class Dealership
 
+  attr_reader :inventory
   def initialize(name, address)
   @name = name
   @address = address
+  @inventory = []
   end
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -6,4 +6,9 @@ class Dealership
   @address = address
   @inventory = []
   end
+
+  def inventory_count
+    @inventory.count
+  end
+
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -5,6 +5,7 @@ class Dealership
   @name = name
   @address = address
   @inventory = []
+  @total_value = 0
   end
 
   def inventory_count
@@ -26,5 +27,16 @@ class Dealership
     end
     end
   end
+
+  def details
+    details = Hash.new { |hash, key| hash[key] }
+    @inventory.each do |car|
+      @total_value += car.total_cost
+      details["total_value"] = @total_value
+      details["address"] = @address
+    end
+     return details
+  end
+
 
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -18,4 +18,13 @@ class Dealership
   def has_inventory?
     @inventory.count > 1
   end
+
+  def cars_by_make(maker)
+    @inventory.each do |car|
+    if  car.make == maker
+      return car
+    end
+    end
+  end
+
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -11,4 +11,7 @@ class Dealership
     @inventory.count
   end
 
+  def add_car(car)
+    @inventory << car
+  end
 end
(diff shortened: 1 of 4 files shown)


lib/dealership.rb
```diff
@@ -0,0 +1,42 @@
...
+    @inventory.count
+  end
+
+  def add_car(car)
+    @inventory << car
+  end
+
+  def has_inventory?
+    @inventory.count > 1
+  end
```
kat3kasper, line 21: This should return `true` if `inventory.count == 1` also so we should be checking that it's greater than 0

```suggestion
  def has_inventory?
    @inventory.count > 0
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,42 @@
+class Dealership
+
+  attr_reader :inventory
+  def initialize(name, address)
+  @name = name
+  @address = address
+  @inventory = []
+  @total_value = 0
```
kat3kasper, line 8: I probably wouldn't save this as an instance variable because you are only calculating and setting the value within the `details` method. Rather I would create a method for it and do the calculation in there.

```
def total_value
  total_value = 0
  @inventory.each do |car|
    total_value += car.total_cost
  end
  total_value
end
```

lib/dealership.rb
```diff
@@ -0,0 +1,42 @@
...
+    @inventory.count > 1
+  end
+
+  def cars_by_make(maker)
+    @inventory.each do |car|
+    if  car.make == maker
+      return car
+    end
+    end
+  end
```
kat3kasper, line 29: You want to be returning an array here rather than 1 car object

```suggestion
  def cars_by_make(maker)
    results = []
    @inventory.each do |car|
      if car.make == maker
        results << car
      end
    end
    results
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def details
+    details = Hash.new { |hash, key| hash[key] }
+    @inventory.each do |car|
+      @total_value += car.total_cost
+      details["total_value"] = @total_value
+      details["address"] = @address
+    end
+     return details
+  end
```
kat3kasper, line 39: If you have a `total_value` method as I showed above then you could simplify this as it doesn't make as much sense to be updating details on each loop.

```suggestion
  def details
    details = Hash.new { |hash, key| hash[key] }
    details["total_value"] = total_value
    details["address"] = @address
    return details
  end
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    expect(dealership.cars_by_make("Toyota")).to eq(car_2)
+    expect(dealership.cars_by_make("Ford")).to eq(car_1)
```
kat3kasper, line 52:
```suggestion
    expect(dealership.cars_by_make("Toyota")).to eq([car_2, car_3])
    expect(dealership.cars_by_make("Ford")).to eq([car_1])
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    expect(dealership.details).to eq({"total_value" => 156000, "address" => "123 Main Street"})
```
kat3kasper, line 65: I think if you would have had a test for `total_value` separately you might have caught your mistake with it's definition.

kat3kasper (commented):
This was a good start Jesse. I appreciate your git commit history. Please keep that up! I would slow down a bit on the specs as you missed a few things there.

PR #48, opened 2021-12-08
Sully Birashk Independent Challenge Week 2

SullyBirashk:


spec/dealership_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    binding.pry
+    expect(dealership.cars_by_make("Ford"))
```
cjsim89, line 57: You were on the right track here - this line finished should've read `.to eq 1`. Try to finish this test out!

lib/dealership.rb
```diff
@@ -0,0 +1,48 @@
...
+  def has_inventory?
+    if @inventory == []
+      return false
+    else
+      return true
+    end
+  end
+
+  def cars_by_make(type)
+    return @inventory.make(type)
```
cjsim89, line 28: I think you were in the middle of this method; see if you can come back to this and finish it out! 
A hint - `#make` is not a method on `@inventory`; you will need to somehow iterate through the `@inventory` to find what you're looking for.

lib/car.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def model
+    @model << @car.split(' ').last
+    return @model[0]
+  end
+
+  def total_cost
+    loan_length * monthly_payment
+  end
+
```
cjsim89, line 25: Missing `#color` and `#paint!` methods from I1.

lib/dealership.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def total_value
+    @total = []
+    @inventory.each do |value|
+    @total << value.total_cost
```
cjsim89, line 34: You were very close here; the `<<` (shovel) operator adds to an array. But we're not expecting an array from the `#total_value` method, we're expecting an Integer. So what gets returned is `[54000]`. The interaction pattern expects an Integer, as so: 
```
pry(main)> dealership.total_value
#=> 156000
```


You were correct in doing a `.each` through the inventory, but all we need is something like: 
`@total += value.total_cost`
and then return that, which you do later. Also, `@total` should not be an array on line 32, it can just be initialized to 0. Finally, it does not need to be a class-level variable with `@`, since it just returns a value from this one method and is not accessed anywhere else in this class file.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    expect(dealership.inventory_count).to eq(2)
+  end
+
+  it 'have inventory?' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    expect(dealership.has_inventory?).to be(false)
```
cjsim89, line 43: This test passes fine, and your code works, this is just a piece of advice: 
When your test is called "has inventory", but you only test for `false`, how can you prove that it **has** inventory? 

You'd need an assertion somewhere (an `expect`) that tests for `true`. That just means adding a car to the Dealership's inventory, and writing an expect has_inventory to be true. Example: 

```
dealership.add_car(car_1)
expect(dealership.has_inventory?).to be true
```

cjsim89: This would ensure you have fully tested the method. But as I said, your logic in the code works out well here, just keep this in mind for the future. :)

lib/dealership.rb
```diff
@@ -0,0 +1,48 @@
...
+    return @total
+    end
+  end
+
+  def details
+    @detals = {}
+    @detals[total_value]
```
cjsim89, line 41: You were close here, remember the syntax for creating & adding to a Hash. 

```
@details = {"total_value" => total_value, "address" => @addy }
```

PR #49, opened 2021-12-08
Nick Cooley

CooleyNick:


lib/dealership.rb
```diff
@@ -0,0 +1,14 @@
...
+  :inventory_count
+  def initialize(dealership, address)
+    @dealership = dealership
+    @address = address
+    @inventory = []
+    @inventory_count = 0
+  end
+
+  def add_car
+    @inventory << @car
```
cjsim89, line 12: `@car` is not a class-level variable, so should not use the `@`. Instead, this method should take an argument of `car`, since in your tests you are passing cars to this method.

lib/car.rb
```diff
@@ -0,0 +1,24 @@
+class Car
+  attr_reader :car, :make, :model, :monthly_payment,
+  :loan_length, :color
+  def initialize(make, model, loan_length)
+    @car = car
+    @make = "Ford"
+    @model = "Mustang"
```
cjsim89, line 7: I would not initialize variables in this way in this instance; they should instead be initalized to the parameters in your `#initialize` method. In other words: 
```
def initialize(make, model, loan_length)
     @make = make
     @model = model
     ...
end
```
Otherwise, you are not actually using the values that you pass in when you call `Car.new("Ford Mustang", 1500, 36)`. In addition, the `model` variable actually is equal to `1500` in this sense - so the order in which you pass in data is very important!

lib/dealership.rb
```diff
@@ -0,0 +1,14 @@
+class Dealership
+  attr_reader :dealership, :address, :inventory,
+  :inventory_count
+  def initialize(dealership, address)
+    @dealership = dealership
```
cjsim89, line 5: No need to initialize the same class in this way; you don't use `@dealership` in this code anywhere, so it's not needed.

lib/car.rb
```diff
@@ -0,0 +1,24 @@
+class Car
+  attr_reader :car, :make, :model, :monthly_payment,
+  :loan_length, :color
+  def initialize(make, model, loan_length)
+    @car = car
```
cjsim89, line 5: No need to initialize the same class in this way; you don't use `@car` in this code anywhere. Only the things that are absolutely necessary for this class to work should be in the `#initialize` method.

lib/car.rb
```diff
@@ -0,0 +1,24 @@
...
+  def total_cost
+    @monthly_payment * @loan_length
+  end
+
+  # def color
+  #   @car = nil
+  # end
+
+  def paint!(color)
+    @color = :blue
```
cjsim89, line 22: You are assigning `@color` to `:blue` here instead of the `color` passed in via argument. Be careful - this may have passed a test you wrote, but if I sent it `:red` instead, it would always equal `:blue`!

spec/car_spec.rb
```diff
@@ -37,4 +37,17 @@
...
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to eq(nil)
+  end
+
+  it 'has been painted blue' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    car.paint!(:blue)
+    expect(car.color).to eq(:blue)
```
cjsim89, line 51: This only passes because `:blue` is the same as in your `Car.paint!` method -- try changing this test to `:red` instead, and you'll see what I mean.

PR #50, opened 2021-12-08
Luke Pascale PC 2 - Dealership

enalihai:
i was almost able to complete iteration 3. I ran into a snag on  cars_by_make that took the last 45 minutes of my time. I feel that I am close to cracking it.

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -19,5 +19,13 @@ class Car
 
   def total_cost
     monthly_payment * loan_length
-  end 
+  end
+
+  def color
+    @color
+  end
+
+  def paint!(color)
+    @color = color
+  end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,23 @@
+class Car
+  attr_reader :name,
+              :monthly_payment,
+              :loan_length
+
+  def initialize(name, monthly_payment, loan_length)
+    @name = name
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+  end
+
+  def make
+    @name = name.split.first
+  end
+
+  def model
+    @name = name.split.last
+  end
+
+  def total_cost
+    monthly_payment * loan_length
+  end 
+end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,7 +1,8 @@
 class Car
   attr_reader :name,
               :monthly_payment,
-              :loan_length
+              :loan_length,
+              :make
 
   def initialize(name, monthly_payment, loan_length)
     @name = name
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,0 +1,30 @@
+class Dealership
+  attr_reader :name,
+              :address
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+  end
+
+  def inventory
+    @inventory
+  end
+
+  def inventory_count
+    inventory.count
+  end
+
+  def add_car(car)
+    @inventory.push(car)
+  end
+
+  def has_inventory?
+    if inventory_count > 0
+     p true
+    else
+     p false
+    end
+  end
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -27,4 +27,15 @@ class Dealership
      p false
     end
   end
+
+  def cars_by_make(make)
+    make.count(make)
+  end
+
+  def total_value
+    inventory_value = []
+    @inventory.each do |total_cost|
+      inventory_value << total_cost
+    end
+  end
 end
--- a/spec/car_spec.rb
+++ b/spec/car_spec.rb
@@ -37,4 +37,24 @@ RSpec.describe Car do
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has a color' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to be(nil)
+  end
+
+#  it 'can paint it' do
+#    car = Car.new("Ford Mustang", 1500, 36)
+#    car.paint!(:blue)
+
+#    expect(car.paint!(:blue)).to be(true)
+#  end
+
+  it 'is a color' do
+    car = Car.new("Ford Mustang", 1500, 36)
+    car.paint!(:blue)
+
+    expect(car.color).to be(:blue)
+  end
 end
--- a/spec/dealership_spec.rb
+++ b/spec/dealership_spec.rb
@@ -61,5 +61,24 @@ RSpec.describe Dealership do
       expect(car_3).to be_a(Car)
       expect(car_4).to be_a(Car)
     end
+
+    it 'can organize by make' do
+      dealership.add_car(car_1)
+      dealership.add_car(car_2)
+      dealership.add_car(car_3)
+      dealership.add_car(car_4)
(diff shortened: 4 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def model
+    @name = name.split.last
+  end
+
+  def total_cost
+    monthly_payment * loan_length
+  end
+
+  def color
```
dcoleman21, line 25: This works. Nice job here. Another way would be to put `@color = nil` in your `initialize` and then add `:color` to your `attr_reader`.

spec/car_spec.rb
```diff
@@ -37,4 +37,24 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has a color' do
```
dcoleman21, line 41: Maybe name this test something like `starts with no color` since in this test we are expecting `nil` when we call `car.color`.

spec/car_spec.rb
```diff
@@ -37,4 +37,24 @@
...
+  end
+
+#  it 'can paint it' do
+#    car = Car.new("Ford Mustang", 1500, 36)
+#    car.paint!(:blue)
+
+#    expect(car.paint!(:blue)).to be(true)
+#  end
+
+  it 'is a color' do
```
dcoleman21, line 54: Then maybe name this test `can get paint` or something like that since in this test we are actually testing for the method `.paint!(color)`. But it will also change the `car.color`. Hope this makes sense.

lib/car.rb
```diff
@@ -0,0 +1,32 @@
+class Car
+  attr_reader :name,
+              :monthly_payment,
+              :loan_length,
+              :make
```
dcoleman21, line 5: You don't need the `:make` here in your `attr_reader` since you created methods below.

lib/car.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize(name, monthly_payment, loan_length)
+    @name = name
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+  end
+
+  def make
+    @name = name.split.first
```
dcoleman21, line 14: Here you would not want to reassign `@name`. Instead you would just call 
```
@name.split.first
```

lib/car.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def model
+    @name = name.split.last
```
dcoleman21, line 18: Same thing here. You would not want to reassign `@name`. Instead you would just call
```
@name.split.last
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      expect(dealership.inventory).to eq([])
+    end
+
+    it 'can count the inventory' do
+
+      expect(dealership.inventory_count).to be(0)
+    end
+
+    it 'has specific cars' do
```
dcoleman21, line 25: You don't need this test in Dealership.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    let(:car_3) {Car.new("Toyota Tercel", 500, 48)}
+    let(:car_4) {Car.new("Chevrolet Bronco", 1250, 24)}
+
+    it 'has no inventory' do
+
+      expect(dealership.has_inventory?).to be(false)
+    end
+
+    it 'has specific cars' do
+
```
dcoleman21, line 58: You don't need this test in Dealership.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      expect(car_4).to be_a(Car)
+    end
+
+    it 'can organize by make' do
+      dealership.add_car(car_1)
+      dealership.add_car(car_2)
+      dealership.add_car(car_3)
+      dealership.add_car(car_4)
+
+      expect(dealership.cars_by_make("Toyota")).to eq(2)
```
dcoleman21, line 71: Pay attention to the interaction pattern. You should have two expects in this test:
```
pry(main)> dealership.cars_by_make("Toyota")
#=> [#<car:0x00007fccd2985f48...>, #<car:0x00007fccd383c2d0...>]

```
This test is expecting  arrays of car objects that match the argument being passed. How would you re-write your expects to implement this?

lib/dealership.rb
```diff
@@ -0,0 +1,41 @@
...
+  attr_reader :name,
+              :address
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+  end
+
+  def inventory
```
dcoleman21, line 11: I would not make this method. You have already put `@inventory = []` in your `initialize` method. Just add `:inventory` to your `attr_reader`.

lib/dealership.rb
```diff
@@ -0,0 +1,41 @@
...
+  def inventory_count
+    inventory.count
+  end
+
+  def add_car(car)
+    @inventory.push(car)
+  end
+
+  def has_inventory?
+    if inventory_count > 0
```
dcoleman21, line 24: Remember that the `>` operator will return a boolean. So you do not need to have this in an `if/else` statement. Just put
```
inventory_count > 0
```

lib/dealership.rb
```diff
@@ -0,0 +1,41 @@
...
+     p true
+    else
+     p false
+    end
+  end
+
+  def cars_by_make(make)
```
dcoleman21, line 31: See the feedback that I left in the `dealership_spec.rb` below for this method. The interaction pattern is expecting an array of car objects that match the argument being passed. This method would then need to use an enumerable to find all of the cars that match.

lib/dealership.rb
```diff
@@ -0,0 +1,41 @@
...
+    make.count(make)
+  end
+
+  def total_value
+    inventory_value = []
+    @inventory.each do |total_cost|
```
dcoleman21, line 37: You are iterating over `@inventory`. 
- What is it that we are wanting to return from this method? 
   - An integer.
- What is `@inventory`? 
    - It is an array of car objects. So your block_variable is a `car object`. You could call it `car`.
- `total_cost` is a method that you created in the `car.rb` class that you can call on the block_variable `car`. Which is what you would want to shovel into your new array to return an integer vs the entire car object.
- Nice job creating an accumulator since you are using a `.each`. 
- Where would you need to return that accumulator in order to have this method give us that new array?
  - After the end for your block

spec/dealership_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    it 'has a total value' do
+
+      expect(dealership.total_value).to eq(156000)
+    end
+
+    it 'can access the dealership details' do
```
dcoleman21, line 79: You wrote this test but might have run out of time to complete the creation of the method. So I would just put an `x` in front of the `it` since you didn't get to it.

dcoleman21 (commented):
Luke, overall nice job putting your tests together and creating methods. I left feedback in specific places where adjustments will be needed. Some things to consider as you continue to get better:
- pay attention to the details of the interaction patterns. Copy/paste exactly what is being asked of you. Specifically with the return values that are expected.
- If you are including things in your attr_reader, then you will not need to create methods as well. If you do not include them in your att_reader, then you will need to create methods. 
Let me know if you have any questions.

PR #51, opened 2021-12-08
TJ Williams 

T-willjr:


lib/car.rb
```diff
@@ -0,0 +1,16 @@
+require 'pry'
+
+class Car
+  attr_reader :car, :monthly_payment, :loan_length, :model, :make
+
+  def initialize(car, monthly_payment, loan_length)
+    @car = car
+    @monthly_payment =  monthly_payment
+    @loan_length = loan_length
+    @make, @model = @car.split(" ")
```
rtillies, line 10: This is a very efficient way to create the `@make` and `@model` attributes. Great work!

lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+    @inventory.count
+  end
+
+  def has_inventory?
+    return true if @inventory.count > 0
+    false
+  end
+
+  def cars_by_make(make)
+    @inventory.select { |car| car if car.make == make }
```
rtillies, line 29: Nice work with using the `select` enumerable, as well as utilizing the shorthand method throughout this class.

lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def details
+    @deets = { "total_value" => total_value,
+               "address" => @address}
+  end
+
+  def average_price_of_car
+    average = total_value / inventory_count
+    average.to_s(:delimited)
```
rtillies, line 45: Well done. Where did you find `:delimited`?

lib/dealership.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def cars_sorted_by_price
+    @inventory.sort_by { |car| car.total_cost }
+  end
+
+  def inventory_hash
+    hash = {"Ford" => cars_by_make("Ford"),
+      "Toyota" => cars_by_make("Toyota"),
+      "Chevrolet" => cars_by_make("Chevrolet") }
+  end
```
rtillies, line 56: Well done using your `cars_by_make` method to create your inventory hash!

lib/car.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def total_cost
+    @monthly_payment * loan_length
+  end
+end
```
rtillies, line 16: The only missing parts to this class is the `@color` attribute and the `paint` method. Did you miss those in the interaction pattern for Iteration 1?
```
pry(main)> car.color
#=> nil

pry(main)> car.paint!(:blue)

pry(main)> car.color
#=> :blue
```

rtillies (commented):
Congrats on finishing Iteration 4! I can tell you are using TDD to write good tests and allowing the tests to write quality code. This is nearly flawless, aside from **completely overlooking** part of Iteration 1. Make sure you're committing after every win (e.g. after every successful test), instead of only at the end of each iteration. Keep learning and keep coding!

rtillies:

lib/car.rb
T-willjr: Yes. I completely missed that section! I'll have to slow down and pay attention next time!

lib/dealership.rb
T-willjr: [StackOverflow!](https://stackoverflow.com/questions/1078347/rails-is-there-a-rails-trick-to-adding-commas-to-large-numbers) I googled it because I figured there had to be a way to return a large number with a comma.

PR #52, opened 2021-12-08
Sam Smith PR car_dealership_2111

samlsmith424:
This is my car dealership ic with parts of I4 attempted.

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,25 @@
+class Car
+  attr_reader :monthly_payment, :loan_length, :color
+  def initialize(make, monthly_payment, loan_length)
+    @make = make
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = nil
+  end
+
+  def make
+    @make.split[0]
+  end
+
+  def model
+    @make.split[1]
+  end
+
+  def total_cost
+    @loan_length * @monthly_payment
+  end
+
+  def paint!(color)
+    @color = color 
+  end
+end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -20,6 +20,6 @@ class Car
   end
 
   def paint!(color)
-    @color = color 
+    @color = color
   end
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,9 +1,17 @@
 class Dealership
-  attr_reader :inventory, :inventory_count
+  attr_reader :inventory
   def initialize(name, location)
     @name = name
     @location = location
     @inventory = []
     @inventory_count = 0
+  end
+
+  def add_car(car)
+    @inventory << car
+  end
+
+  def inventory_count
+    @inventory.size
   end
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,0 +1,9 @@
+class Dealership
+  attr_reader :inventory, :inventory_count
+  def initialize(name, location)
+    @name = name
+    @location = location
+    @inventory = []
+    @inventory_count = 0
+  end
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,10 +1,11 @@
 class Dealership
   attr_reader :inventory
-  def initialize(name, location)
+  def initialize(name, address)
     @name = name
-    @location = location
+    @address = address
     @inventory = []
     @inventory_count = 0
+    @has_inventory = false
   end
 
   def add_car(car)
@@ -13,5 +14,41 @@ class Dealership
 
   def inventory_count
     @inventory.size
+  end
+
+  def has_inventory?
+    @has_inventory
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |car|
+      car.make == make
+    end
+  end
+
+  def total_value
+    total_v = []
+    @inventory.each do |car|
+      total_v << car.total_cost
+    end
+    total_v.reduce(0, :+)
+  end
+
+  def details
+    details = {}
+    details[(:total_value).to_s] = 156000
+    details[(:address).to_s] = "123 Main Street"
+    details
+  end
+
+  def average_price_of_car
+    (total_value / 4).to_s
+  end
+
+  def cars_sorted_by_price
+
+    # @inventory.map do |car|
+    #   car.total_cost
+    # end.sort
   end
 end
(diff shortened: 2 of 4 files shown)


spec/car_spec.rb
```diff
@@ -37,4 +37,12 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'is does not have a color by unless painted' do
+    car = Car.new("Ford Mustang", 1500, 36)
```
dcoleman21, line 42: This test looks great.

lib/car.rb
```diff
@@ -0,0 +1,25 @@
+class Car
+  attr_reader :monthly_payment, :loan_length, :color
```
dcoleman21, line 2: This all looks great.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,74 @@
+require './lib/car'
+require './lib/dealership'
+
+RSpec.describe Dealership do
+
```
dcoleman21, line 5: I would add an `it exisits` test to test that a `dealership` is and instance of `Dealership`.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.total_value).to eq(156000)
+    expect(dealership.details).to eq({"total_value" => 156000, "address" => "123 Main Street"})
```
dcoleman21, line 57: You might put this one into its own test. (`.details`)

spec/dealership_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.average_price_of_car).to eq("39,000")
+    expect(dealership.cars_sorted_by_price).to eq([car_3, car_4, car_2, car_1])
```
dcoleman21, line 72: I would also put this into its own test. (`.cars_sorted_by_price`)

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
+class Dealership
+  attr_reader :inventory
```
dcoleman21, line 2: You will want to also have 
```
:name,
:address
```
in your `attr_reader`

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+    @inventory_count = 0
```
dcoleman21, line 7: This is a method that you created below. You do not need this here.

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+    @has_inventory = false
```
dcoleman21, line 8: You will not want to have this here either. You will do the logic in your `has_inventory?` method below.

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+  def add_car(car)
+    @inventory << car
+  end
+
+  def inventory_count
+    @inventory.size
+  end
+
+  def has_inventory?
+    @has_inventory
```
dcoleman21, line 20: So we are getting our inventory count in the method above. So here, how would we know if we had inventory? What if we checked that our `inventory_count > 0`??

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |car|
```
dcoleman21, line 24: Nice job using `.select` here since we want to find all of the cars that meet the criteria passed to the method.

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+      car.make == make
+    end
+  end
+
+  def total_value
+    total_v = []
+    @inventory.each do |car|
```
dcoleman21, line 31: Nice job solving this with `.each` and then using `.reduce` to return an integer. Take a look at the `.sum` enumerable. 
https://ruby-doc.org/core-2.4.1/Array.html#:~:text=sum-,(,-init%3D0)%20%E2%86%92%20number

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+      total_v << car.total_cost
+    end
+    total_v.reduce(0, :+)
+  end
+
+  def details
+    details = {}
```
dcoleman21, line 38: So very close here. A couple things:
1. You want your keys to be strings, which you are doing with the `.to_s` but you could just do `details["total_value"]`
2. You are what we call `hard coding` the values. So they pass based on the test you wrote but we want it to be dynamic. Is there a method for `total_value`? And do you have your `@address` in the initialize? That is what you would set you keys equal to. 
```
details["total_value"] = total_value
details["address"] = @address
```

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+    details[(:total_value).to_s] = 156000
+    details[(:address).to_s] = "123 Main Street"
+    details
+  end
+
+  def average_price_of_car
+    (total_value / 4).to_s
```
dcoleman21, line 45: Again super close but you are hard coding with the 4. 
1. Where do we get the number of things in our inventory? (`inventory_count`)
2. We do want it as a string but there is also a comma we have to get in there. Check out `.insert`. 
https://ruby-doc.org/core-2.4.1/Array.html#:~:text=Inserts%20the%20given-,values,-before%20the%20element

lib/dealership.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def cars_sorted_by_price
+
```
dcoleman21, line 49: Check out `.sort_by` for this one. 
https://ruby-doc.org/core-2.4.1/Enumerable.html#:~:text=comparison%20is%20expensive.-,sort_by,-%7B%20%7Cobj%7C%20block%20%7D%20%E2%86%92%20array

dcoleman21 (commented):
Overall Sam really nice job! I have left feedback for you to review. Let me know if you have any questions.

PR #53, opened 2021-12-08
Adrian Campos car_dealership_2111

adriancampos29:


lib/car.rb
```diff
@@ -0,0 +1,25 @@
+class Car
+  attr_reader :make, :model, :monthly_payment
```
kat3kasper, line 2: Since you made methods for `make` and `model` below you wouldn't add them to attr_reader. You only need to add attributes that don't have their own method. `monthly_payment` for example is good! You defined it in the initialize and then since you added it to attr_reader you can access it on a car. 

You should do the same thing for loan_length rather than creating a method for it like you are doing on line 17. 

```suggestion
  attr_reader :monthly_payment, :loan_length
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,14 @@
+require './lib/car'
+require './lib/dealership'
+
+RSpec.describe Dealership do
+  it 'exists' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    expect(dealership.inventory). to eq([])
+  end
```
kat3kasper, line 8: Since you are tesing that "it exists", I would only test that the dealership is a correct type. Then I would test if it had an inventory separately:
```suggestion
  it 'exists' do
    dealership = Dealership.new("Acme Auto", "123 Main Street")
    expect(dealership).to be_a Dealership
  end
  
   it 'has an inventory' do
    dealership = Dealership.new("Acme Auto", "123 Main Street")
    expect(dealership.inventory).to eq([])
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,13 @@
...
+  attr_reader :inventory
+  def initialize(dealership, address)
+    @dealership = dealership
+    @address = address
+    @inventory = []
+  end
+
+  def inventory_count
+    0
+  end
```
kat3kasper, line 11: This is a start, but I would calculate the inventory_count from @inventory so:
```suggestion
  def inventory_count
    @inventory.length
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,13 @@
+class Dealership
+  attr_reader :inventory
+  def initialize(dealership, address)
+    @dealership = dealership
+    @address = address
+    @inventory = []
+  end
```
kat3kasper, line 7: It could be a little confusing to use `dealership` as a variable here. I might use `name` instead
```suggestion
  def initialize(name, address)
    @name = name
    @address = address
    @inventory = []
  end
```

lib/car.rb
```diff
@@ -0,0 +1,25 @@
...
+    @loan_length = loan_length
+  end
+
+  def make
+    @car.split.first
+  end
+
+  def model
+    @car.split.last
+  end
```
kat3kasper, line 15: Good job with these!

kat3kasper (commented):
Good starting effort Adrian. I really appreciated your git history with numerous commits to follow your process. Definitely keep that up! Unfortunately, you weren't able to get too far in the time allotted. I would practice object creation more so that you are able to get farther a bit faster

PR #56, opened 2021-12-08
Gunnar Sorensen

glsorensen:


lib/car.rb
```diff
@@ -0,0 +1,26 @@
+require 'pry'
+class Car
+  attr_reader :name,
+              :make,
+              :model,
```
kat3kasper, line 5: Since you are defining methods for make (line 15) and model (line 19), it doesn't make sense to also declare them with `attr_reader`. `attr_reader` is used when you when you want to create "getter" methods on attributes that will be accessed outside of the Car class but we don't want to manually write out the method. So for `monthly_payment` it works because ruby is creating the following method for us:

```
def monthly_payment
  @monthly_payment
end
```

So since you are defining `make` and `model` already you don't need them to be created by ruby.

lib/car.rb
```diff
@@ -0,0 +1,26 @@
+require 'pry'
+class Car
```
kat3kasper, line 2: You are missing `color` attribute and `paint!` method defined in Iteration 1 spec.

lib/dealership.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def add_car(car)
+    @inventory.push(car)
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |invent|
+      invent.cars.make == make
+    end
+  end
```
kat3kasper, line 24: This is close. When you call `.select` you are iterating through each object within the @inventory array. You have access to that object via the variable defined between `| |` so for you that is `invent` and `invent` is the Car object. You can then call make on the car:

```suggestion
  def cars_by_make(make)
    @inventory.select do |invent|
      invent.make == make
    end
  end
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,62 @@
+require './lib/car'
+require './lib/dealership'
+
+
+RSpec.describe Car do
+  context 'Iteration II' do
+    let (:dealership) {Dealership.new("Acme Auto", "123 Main Street")}
+    let (:car_1) {Car.new("Ford Mustang", 1500, 36)}
+    let (:car_2) {Car.new("Toyota Prius", 1000, 48)}
```
kat3kasper, line 9: Like the use of `let` here to set up all of your tests so that you don't need to repeat yourself

spec/dealership_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      dealership.add_car(car_2)
+
+      expect(dealership.inventory).to eq([car_1, car_2])
+      expect(dealership.inventory_count).to eq(2)
+    end
+  end
+
+
+
+  context 'Iteration III' do
```
kat3kasper, line 38: Nice job with TDDing! You are missing `has_inventory?` from iteration 3.

kat3kasper (commented):
This is a good start Gunnar. I'd like to see a few more commits while you build out your code so that you are chunking smaller increments. Good work with following through with TDDing. I'd slow down a bit on reading through the spec, because you missed some of the functionality we were looking for `color`, `paint!` and `has_intentory?`. Keep practicing with object creation so you can get through this a bit faster and get through more iterations.

PR #58, opened 2021-12-08
Aedan Yturralde

aedanjames:
I was able to get through Iteration 3, but had trouble in iteration 4 converting the average price of the car to include a comma in 39,000. Tried messing with number_with_deliminator method but I think my syntax was lacking.

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,29 @@
+require 'pry'
+class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length, :color
+  def initialize(make_model, monthly_payment, loan_length)
+    @make_model = make_model
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = color
+  end
+
+  def make
+    make = @make_model.split()
+    return make.first
+  end
+
+  def model
+    model = @make_model.split()
+    return model.last
+  end
+
+  def total_cost
+    monthly_payment * loan_length
+  end
+
+  def paint!(color)
+    @color = color
+  end
+
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,3 +1,4 @@
+require 'pry'
 class Dealership
   attr_reader :inventory
   def initialize(argument_1, argument_2)
@@ -17,5 +18,23 @@ class Dealership
   def has_inventory?
     @inventory = []
       false
+  end
+
+  def cars_by_make(make)
+    car_make = []
+    @inventory.each do |car|
+      if car.make == make
+        car_make << car
+      end
+    end
+    car_make
+  end
+
+  def total_value
+    total = []
+    @inventory.each do |car|
+      total << car.total_cost
+    end
+    total.sum
   end
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -32,9 +32,13 @@ class Dealership
 
   def total_value
     total = []
-    @inventory.each do |car|
+    @inventory.find_all do |car|
       total << car.total_cost
     end
     total.sum
+  end
+
+  def average_price_of_car
+    (total_value / (@inventory.size))
   end
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,0 +1,6 @@
+class Dealership
+  def initialize(argument_1, argument_2)
+    @argument_1 = argument_1
+    @argument_2 = argument_2
+  end
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,6 +1,12 @@
 class Dealership
+  attr_reader :inventory
   def initialize(argument_1, argument_2)
     @argument_1 = argument_1
     @argument_2 = argument_2
+    @inventory = []
+  end
+
+  def inventory_count
+    @inventory.count
   end
 end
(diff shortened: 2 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,29 @@
+require 'pry'
```
kat3kasper, line 1: Try to remove `pry` before summitting

lib/car.rb
```diff
@@ -0,0 +1,29 @@
...
+class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length, :color
```
kat3kasper, line 3: Since you have defined methods for make and model below, it is not necessary to also add them to the attr_reader list. attr_reader essentially creates a hidden "getter" method for any attributes in the list. For example with :monthly_payment ruby creates this method for us:

```
def monthly_payment
  @monthly_payment
end
```

but because you have the methods already defined it doesn't make sense

```
  def make
    make = @make_model.split()
    return make.first
  end
```

```suggestion
  attr_reader :monthly_payment, :loan_length, :color
```

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
+require 'pry'
+class Dealership
+  attr_reader :inventory
+  def initialize(argument_1, argument_2)
+    @argument_1 = argument_1
+    @argument_2 = argument_2
+    @inventory = []
+  end
```
kat3kasper, line 8: I might rename `argument_1` and `argument_2` to `name` and `address`
```suggestion
  def initialize(name, address)
    @name = name
    @address = address
    @inventory = []
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def inventory_count
+    @inventory.count
```
kat3kasper, line 11: 🙌 Loved that you used .count here on @inventory

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def add_car(car)
+    @inventory << car
+  end
+
+  def has_inventory?
+    @inventory = []
+      false
+  end
```
kat3kasper, line 21: I think you wanted to check if @inventory was empty

```
def has_inventory?
  if (@inventory == [])
    return false
  end
  true
end
```

Another way to do it:
```suggestion
  def has_inventory?
    inventory_count > 0
  end
```

spec/car_spec.rb
```diff
@@ -37,4 +37,14 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has a color attribute' do
+    car = Car.new("Ford Mustang", 1500, 36)
+
+    expect(car.color).to eq(nil)
+    car.paint!(:blue).equal? (:blue)
+    expect(car.color).to eq(:blue)
```
kat3kasper, line 46: Good before and after checks

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def cars_by_make(make)
+    car_make = []
+    @inventory.each do |car|
+      if car.make == make
+        car_make << car
+      end
+    end
+    car_make
+  end
```
kat3kasper, line 31: This works, but can you think of a way to refactor this to use `.select`?

spec/dealership_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  it 'Has inventory if it has more than one car' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    expect(dealership.has_inventory?).to eq(false)
+  end
```
kat3kasper, line 37: I think you would have caught your error if you set up this test correctly. You need to add the cars to the delearship

```suggestion
  it 'Has inventory if it has more than one car' do
    dealership = Dealership.new("Acme Auto", "123 Main Street")
    expect(dealership.has_inventory?).to eq(false)
    
    car_1 = Car.new("Ford Mustang", 1500, 36)
    car_2 = Car.new("Toyota Prius", 1000, 48)
    car_3 = Car.new("Toyota Tercel", 500, 48)
    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
    dealership.add_car(car_1)
    dealership.add_car(car_2)
    dealership.add_car(car_3)
    dealership.add_car(car_4)
    
    expect(dealership.has_inventory?).to eq(true)
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,44 @@
...
+  def total_value
+    total = []
+    @inventory.find_all do |car|
+      total << car.total_cost
+    end
+    total.sum
+  end
+
+  def average_price_of_car
+    (total_value / (@inventory.size))
```
kat3kasper, line 42: You could have done something like this to insert a comma:
```suggestion
    (total_value / (@inventory.size)).to_s.insert(2, ",")
```

kat3kasper (commented):
Nice start Aedan. Love your commits along the way keep that up!

You missed `.details` on Iteration 3.

PR #59, opened 2021-12-08
Austin M. car_dealership challenge

AustinCMoore:
This is Austin Moore's attempt at the IC for car dealership

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,36 @@
+class Car
+  attr_reader :name,
+              :monthly_payment,
+              :loan_length,
+              :make,
+              :model,
+              :total_cost,
+              :color
+
+  def initialize(name, monthly_payment, loan_length) #can i split name into make/model here?
+    @name            = name
+    @monthly_payment = monthly_payment
+    @loan_length     = loan_length
+    @make            = make
+    @model           = model
+    @total_cost      = total_cost
+    @color          = :color
+  end
+
+  def make
+    @make = @name.split.first
+  end
+
+  def model
+    @model = @name.split.last
+  end
+
+  def total_cost
+    @total_cost = monthly_payment * loan_length
+  end
+
+  def paint!(paint)
+    @color = paint
+  end
+
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -50,4 +50,37 @@ class Dealership
     }
   end
 
+  def average_price_of_car #refactor - how to get value from total_value
+    temp_value = 0
+    @inventory.each do |car| #this is bloated
+      temp_value += car.total_cost
+    end
+    average = temp_value / @inventory.length #needs to convert .to_s after inputting a comma
+    if average.to_s.chars.to_a.length >= 4 #if over 3, string needs commas
+      average.to_s.chars.to_a #create array
+      #combine every 3 characters
+      #input commas between groups of 3
+      #output whole string with commas in correct order
+    else #if not over 3 no commas necessary
+      average.to_s
+    end
+  end
+
+  def cars_sorted_by_price
+    #for each car
+    #return car.total_cost & save to array
+    #Does a sort method exist for an array? If not create comparisons & sort
+    #output
+  end
+
+  def inventory_hash
+    #create new empty hash
+    #for each car
+    #if no key exists for a car.make, then create a key with car.make
+    #if key exists, skip
+    #after each key is created, for each car
+    #If car.make == key, shovel car instance into array
+    #return hash
+  end
+
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,0 +1,18 @@
+class Dealership
+  attr_reader :dealer_name,
+              :address,
+              :inventory,
+              :inventory_count
+
+  def initialize(dealer_name, address)
+    @dealer_name      = dealer_name
+    @address          = address
+    @inventory        = []
+    @inventory_count  = 0
+  end
+
+  def add_car(car)
+    @inventory << car
+    @inventory_count += 1
+  end
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -2,17 +2,52 @@ class Dealership
   attr_reader :dealer_name,
               :address,
               :inventory,
-              :inventory_count
+              :inventory_count,
+              :value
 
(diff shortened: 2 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def initialize(name, monthly_payment, loan_length) #can i split name into make/model here?
+    @name            = name
+    @monthly_payment = monthly_payment
+    @loan_length     = loan_length
+    @make            = make
+    @model           = model
+    @total_cost      = total_cost
+    @color          = :color
+  end
```
kat3kasper, line 18: Yup you could split make/model directly in the initialize:

```
def initialize(name, monthly_payment, loan_length) 
  ...
  @make            = name.split.first
  @model           = name.split.last
end
```

I personally wouldn't put total cost in the initialize and would leave it as a method below that can still be called on the object at anytime. It would also not be needed in attr_reader because you are defining a method

```
def total_cost
  @monthly_payment * @loan_length
end
```

In the specs `@color` should be initialized to `nil`

```
def initialize(name, monthly_payment, loan_length) 
  ...
  @color = color
end
```

lib/dealership.rb
```diff
@@ -0,0 +1,86 @@
...
+    @inventory_count += 1
+  end
+
+  def has_inventory?
+    if @inventory_count = 0
+      return false
+    else
+      return true
+    end
+  end
```
kat3kasper, line 27: You are doing a variable assignment here rather than checking if it is equal

```suggestion
  def has_inventory?
    if @inventory_count == 0
      return false
    else
      return true
    end
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,86 @@
...
+              :inventory,
+              :inventory_count,
+              :value
+
+  def initialize(dealer_name, address)
+    @dealer_name      = dealer_name
+    @address          = address
+    @inventory        = []
+    @inventory_count  = 0
+    @value            = 0
```
kat3kasper, line 13: I think `@value` is messing you up here. We don't need to set it as an instance variable and should only worry about it when calculating the total value.

lib/dealership.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+    @value
+  end
+
+  def details
+    detail = {
+      "total_value" => @value,
+      "address"     => @address
+    }
+  end
```
kat3kasper, line 51: Close. I would set this to the result from the `total_value` method

```suggestion
  def details
    detail = {
      "total_value" => total_value,
      "address"     => @address
    }
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+    car_makes
+  end
+
+  def total_value
+    @inventory.each do |car|
+      @value += car.total_cost
+    end
+    @value
+  end
```
kat3kasper, line 44: You want to be using a local variable here rather than an instance variable. 

```suggestion
  def total_value
    value = 0
    @inventory.each do |car|
      value += car.total_cost
    end
    value
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,86 @@
...
+    average = temp_value / @inventory.length #needs to convert .to_s after inputting a comma
+    if average.to_s.chars.to_a.length >= 4 #if over 3, string needs commas
+      average.to_s.chars.to_a #create array
+      #combine every 3 characters
+      #input commas between groups of 3
+      #output whole string with commas in correct order
+    else #if not over 3 no commas necessary
+      average.to_s
+    end
+  end
```
kat3kasper, line 67: I really like the use of psuedocode here! I think this is a great way to approach building out methods

spec/dealership_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+
+    expect(dealership.inventory).to eq([car_1, car_2])
+    expect(dealership.inventory_count).to eq(2)
+  end
+
+  it 'has inventory?' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+
+    expect(dealership.has_inventory?).to eq(false)
+  end
```
kat3kasper, line 34: You did the empty check but not the check of `has_inventory?` after adding some cars to the Dealership

spec/dealership_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    dealership.total_value
```
kat3kasper, line 75: This only passed because you called `dealership.total_value` before calling `dealership.details`. This shouldn't be required however. You should be able to call `dealership.details` and it return the `correct total_value`

spec/dealership_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+    car_2 = Car.new("Toyota Prius", 1000, 48) #48000
+    car_3 = Car.new("Toyota Tercel", 500, 48) #24000
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24) #30000
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.inventory_hash).to eq({"Ford" => [car_1], "Toyota" => [car_2, car_3], "Chevrolet" => [car_4]})
+  end
```
kat3kasper, line 120: Love the TDD here!

kat3kasper (commented):
Good start on things Austin. You missed a few things from the spec that I think you may have caught if you slowed down a bit like `color` and  `has_inventory?`. Really awesome job TDDing throughout your challenge and good work with many detailed git commits; definitely keep that up! I would suggest you review some lessons on variable scoping and object creation as it should help a bit with your use of variables and some of the issues I saw when reviewing your code. 

https://backend.turing.edu/module1/lessons/scope

PR #60, opened 2021-12-08
Tyler Ross

Tross0208:


spec/car_spec.rb
```diff
@@ -37,4 +37,18 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has color' do
```
dcoleman21, line 41: Nice job on this test.

lib/car.rb
```diff
@@ -0,0 +1,22 @@
+class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length, :color
+  def initialize(name, monthly_payment, loan_length)
+    @name = name
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color
+
+    split = name.split(' ')
```
dcoleman21, line 9: Nice job getting this. This might be better to make into a method. So something like:
```
def name_split
   @name.split(' ')
end
```
Then in your initialize you could update `@make` and `@model` to:
```
@make = name_split.first
@model = name_split.last
```
Essentially what you have done in the initialize but in a method instead to clean up your initialize.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  it 'exists' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+
+    expect(dealership).to be_a Dealership
+  end
+
+  it 'can have inventory' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    dealership.inventory
```
dcoleman21, line 16: You do not need this here. This is what you are calling on line 18

spec/dealership_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    dealership.add_car(car_2)
+
+    expect(dealership.inventory_count).to eq(2)
+  end
+
+  it 'has multiple cars in inventory' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    dealership.add_car(car_1)
+
```
dcoleman21, line 51: You expect on line 53 would actually return `true` because you have added a car on line 50. You would need to write another expect on line 49.5 in order to have it return `false`.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    dealership.add_car(car_1)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+
+    expect(dealership.cars_by_make("Toyota")).to eq([car_2, car_3])
```
dcoleman21, line 68: There is another expect that the interaction pattern asked for 
```
```

lib/dealership.rb
```diff
@@ -0,0 +1,47 @@
+class Dealership
+  attr_reader :name, :address, :inventory
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+    @inventory_count = 0
```
dcoleman21, line 8: You do not need this here. `inventory_count` is not a counter, it is a method.

lib/dealership.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def add_car(car)
+    @inventory << car
+  end
+
+  def inventory_count
+    @inventory_count = @inventory.length
```
dcoleman21, line 16: You do not need to create this extra instance variable called `@inventory_count`. This method will return that for your with the `inventory.length`.

lib/dealership.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def has_inventory?
+    if @inventory_count < 2
```
dcoleman21, line 20: How many cars would our `inventory_count` need to have in it to return `false`? If we had 0 cars would we have any inventory? So what would we want `inventory_count` to be greater than?

lib/dealership.rb
```diff
@@ -0,0 +1,47 @@
...
+      return false
+    end
+  end
+
+  def cars_by_make(make_check)
+    @cars_by_make = []
```
dcoleman21, line 26: You do not need the `@` before this accumulator. It will only be scoped to this method.

lib/dealership.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    @inventory.each do |car|
+      if car.make == make_check
+        @cars_by_make << car
+      end
+    end
+    return @cars_by_make
```
dcoleman21, line 33: Nice job making sure that your accumulator is the last line of code in this method so that we can have our changed array. 

Challenge for you. What are we wanting here? To find all of the cars that match the argument, right? Can you think of an enumerable the will do that?

lib/dealership.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def total_value
+    value=0
+
+    @inventory.each do |car|
```
dcoleman21, line 39: Nice job using the `.each` here with your counter. Look into the `.sum` enumerable. Since what we are wanting to do is return the sum of our array.
https://ruby-doc.org/core-2.4.1/Enumerable.html#:~:text=sum-,(,-init%3D0)%20%E2%86%92%20number

dcoleman21 (commented):
Overall Tyler nice job. Looks like you are getting the hang of following the interaction pattern. Keep practicing and make sure to pay attention to the details. I have left feedback for you. Let me know if you have any questions.

PR #61, opened 2021-12-08
Becky Nisttahuz Car dealership IC

benistta:


lib/car.rb
```diff
@@ -0,0 +1,36 @@
+class Car
+  attr_reader :make, :model, :payments, :loan, :color
+
+  def initialize(type, monthly_payment, loan_length)
+    @make = type.split.first
+    @payments = monthly_payment
+    @loan = loan_length
```
kat3kasper, line 7: I would have used `loan_length` rather than `loan` and `monthly_payment` rather than `payments` as the instance variable as well. That was you wouldn't have to create the methods (monthly_payment line 14  or loan_length line 18) below.
```
attr_reader :loan_length, :monthly_payment

def initialize(type, monthly_payment, loan_length)
  ...
  @loan_length = loan_length
  @monthly_payment = monthy_payment
end
```

lib/dealership.rb
```diff
@@ -0,0 +1,43 @@
...
+    if @inventory.count <= 0
+      false
+    elsif @inventory.count > 0
+      true
+    end
+  end
+
+  def inventory_count
+    @inventory.count
+  end
```
kat3kasper, line 20: Loved that you used the `@inventory` to find the count!

lib/dealership.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def add_car(car)
+    @inventory << car
+  end
+
+  def cars_by_make(make)
+    make_cars = []
+    @cars.each do |car|
```
kat3kasper, line 28: All of your cars are being stored in `@inventory` so if you change @cars to @inventory this would work!
```suggestion
    @inventory.each do |car|
```

lib/dealership.rb
```diff
@@ -0,0 +1,43 @@
...
+      if car.make == make
+        make_cars << car
+      end
+    end
+    return make_cars
+  end
+
+  def total_value
+    @inventory.@payment.count
```
kat3kasper, line 37: this broke your code, unfortunately. If you run into that problem it could be helpful to comment out until you figure out how to get it working

spec/car_spec.rb
```diff
@@ -37,4 +37,16 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it 'has a total cost' do
```
kat3kasper, line 41: You want to make sure you are updating the tests' names when creating new ones rather than keeping what's left on copy & paste
```suggestion
  it 'has a color' do
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  it 'has an inventory count with two cars' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    expect(dealership.inventory).to eq([car_1, car_2])
+    expect(dealership.inventory_count).to eq(2)
+  end
```
kat3kasper, line 31: Nice job with the testing here!

kat3kasper (commented):
This is a good start Becky. I'd keep practicing object creation so you can get a bit faster and get through more iterations in the future. 

Keep up the good work on the numerous commits throughout your work!

PR #62, opened 2021-12-08
Joe Galvin

jwgalvin:
got through most of 4.  I have trouble with accessing hashes.

joe

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -16,4 +16,8 @@ class Car
     model = name[0].split(" ")
     model.last
   end
+
+  def total_cost
+    total = monthly_payment * loan_length
+  end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -3,13 +3,17 @@ class Car
 
   def initialize(name, payment, loan_length)
     @name = name,
-    @payment = payment,
+    @monthly_payment = payment,
     @loan_length = loan_length
   end
 
   def make
-    make_temp = []
     make = name[0].split(" ")
     make.first
+  end
+
+  def model
+    model = name[0].split(" ")
+    model.last
   end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -5,6 +5,7 @@ class Car
     @name = name,
     @monthly_payment = payment,
     @loan_length = loan_length
+    @total
   end
 
   def make
@@ -18,6 +19,6 @@ class Car
   end
 
   def total_cost
-    total = monthly_payment * loan_length
+    @total = monthly_payment * loan_length
   end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,15 @@
+class Car
+  attr_reader :name, :monthly_payment, :loan_length
+
+  def initialize(name, payment, loan_length)
+    @name = name,
+    @payment = payment,
+    @loan_length = loan_length
+  end
+
+  def make
+    make_temp = []
+    make = name[0].split(" ")
+    make.first
+  end
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -14,4 +14,12 @@ class Dealership
   def add_car(car)
     @inventory << car
   end
+
+  def has_inventory?
+    if @inventory.length == 0
+      false
+    else
+      true
+    end
+  end
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,0 +1,14 @@
+class Dealership
+  attr_reader :name, :address, :inventory
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+  end
+
+  def inventory_count
+    @inventory.count
+  end
+
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -32,4 +32,13 @@ class Dealership
     end
     car_makes
   end
+
+  def total_value
+    tot_value = 0
+    @inventory.each do |car|
+      tot_value += car.total_cost
+    end
+    tot_value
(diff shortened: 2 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,24 @@
+class Car
```
kat3kasper, line 1: You missed `color` and `paint!` as defined in Iteration 1 spec.

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
+class Dealership
+  attr_reader :name, :address, :inventory
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+    @total_value = 0
```
kat3kasper, line 8: You don't need `@total_value` here because you are creating a method to do the calculation later on

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
...
+    car_makes
+  end
+
+  def total_value
+    @tot_value = 0
+    @inventory.each do |car|
+      @tot_value += car.total_cost
+    end
+    return @tot_value
+  end
```
kat3kasper, line 44: Rather than using an instance variable `@tot_value` I would use a local variable to do the calculation

```suggestion
  def total_value
    tot_value = 0
    @inventory.each do |car|
      tot_value += car.total_cost
    end
    return tot_value
  end
```

Then everywhere in the code you would call `total_value`. 

If you do it this way and you try to call `average_price_of_car` or `details` on dealership before you every call `total_value` you will get @tot_value == nil

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def details
+    deats = Hash.new
+    deats = { @tot_value => @address }
+  end
```
kat3kasper, line 49:
```suggestion
  def details
    deats = Hash.new
    deats = { "total_value" =>  total_value, "address" => @address }
  end
```

lib/car.rb
```diff
@@ -0,0 +1,24 @@
...
+  attr_reader :name, :monthly_payment, :loan_length
+
+  def initialize(name, payment, loan_length)
+    @name = name,
+    @monthly_payment = payment,
+    @loan_length = loan_length
+    @total
```
kat3kasper, line 8: I'm not sure what `@total` is here

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    # @inventory.each do |car|
+    #   avg_price += car.total_cost
+    #    avg_price= avg_price/@inventory.count
+    # end
+    avg_price.to_s
+  end
+
+  def cars_sorted_by_price
+    @inventory.sort_by(&:monthly_payment)
```
kat3kasper, line 62: We want to sort by `total_cost` not only the `monthly_payment`

lib/dealership.rb
```diff
@@ -0,0 +1,64 @@
...
+    return @tot_value
+  end
+
+  def details
+    deats = Hash.new
+    deats = { @tot_value => @address }
+  end
+
+  def average_price_of_car
+    avg_price = @tot_value / @inventory.count
```
kat3kasper, line 52:
```suggestion
    avg_price = total_value / @inventory.count
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    expect(dealership.inventory).to eq([car_1, car_2])
+    expect(dealership.inventory_count).to eq(2)
+  end
+
+  it 'IT3 p1' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    dealership.has_inventory?
+    expect(dealership.has_inventory?).to eq(false)
```
kat3kasper, line 25: I would have like to see that you also checked that this returned `true` after you added some cars

spec/dealership_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    expect(dealership.cars_by_make("Toyota")).to eq([car_2, car_3])
+    expect(dealership.cars_by_make("Ford")).to eq([car_1])
+    expect(dealership.total_value).to eq(156000)
+    #expect(dealership.details).to eq({15600, "123 Main Street"})
+  end
+
+  it 'IT4 p1' do
```
kat3kasper, line 44: Before moving to Iteration 4 I would have like to see the test for `.details` even if you couldn't get it to pass

spec/dealership_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    dealership.total_value
+    expect(dealership.average_price_of_car).to eq("39000")
+    expect(dealership.cars_sorted_by_price).to eq([car_3, car_2, car_4, car_1])
```
kat3kasper, line 56: Unfortunately, this isn't the correct order as we want to be sorting on total cost so it should be [car_3, car_4, car_2, car_1]

kat3kasper (commented):
This is a good start Joe, but I would slow down a bit so that to make sure you complete everything in each iteration. You missed `color`, `paint!` and the test for `details`. Nice job on the git commit history; keep that up!

PR #63, opened 2021-12-08
Ian Ross

ross-ian28:
Unfortunately I had to start an hour late because I was busy throwing up, but I made it as far as I could!

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -17,4 +17,8 @@ class Car
     return @name[1]
   end
 
+  def total_cost
+    return loan_length * monthly_payment
+  end
+
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -11,4 +11,10 @@ class Car
     @name = @name.split(" ")
     return @name[0]
   end
+
+  def model
+    @name = @name.split(" ")
+    return @name[1]
+  end
+
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -8,7 +8,7 @@ class Car
   end
 
   def make
-    @name.split(" ")
-    return @name
+    @name = @name.split(" ")
+    return @name[0]
   end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -21,4 +21,11 @@ class Car
     return loan_length * monthly_payment
   end
 
+  # def color
+  #   @color = nil
+  # end
+  #
+  # def paint!
+  # end
+
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,14 @@
+class Car
+  attr_reader :name, :monthly_payment, :loan_length
+
+  def initialize(name, monthly_payment, loan_length)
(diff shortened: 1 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def make
+    @name = @name.split(" ")
+    return @name[0]
+  end
+
+  def model
+    @name = @name.split(" ")
+    return @name[1]
+  end
```
kat3kasper, line 18: The problem with these are that you are overwriting the instance variable @name when `make` or `model` are called. It would make more sense to use local variables that scoped only to these methods:

```
def make
  name = @name.split(" ")
  return name[0]
end
```

or even using a different variable name if that is more clear

```
def make
  make_model = @name.split(" ")
  return make_model[0]
end
```

lib/car.rb
```diff
@@ -0,0 +1,31 @@
...
+  def total_cost
+    return loan_length * monthly_payment
+  end
+
+  # def color
+  #   @color = nil
+  # end
+  #
+  # def paint!
+  # end
```
kat3kasper, line 29: appreciate that you started these. What I would do for color is to initialize it to nil within the initializer so when you call `.color` before painting it returns `nil`

```
def initialize(name, monthly_payment, loan_length)
    @name = name
    ....
    @color = nil
  end
```

then you can declare the `paint!` function to take in 1 parameter and update your instance variable:

```
def paint!(color)
  @color = color
end
```

lib/dealership.rb
```diff
@@ -0,0 +1,16 @@
...
+  attr_reader :name, :location
+
+  def initialize(name, location)
+    @name = name
+    @location = location
+  end
+
+  def inventory
+    @inventory = []
+  end
```
kat3kasper, line 11: You want to set this during the initialization that way you can update @inventory later otherwise it will only every return `[]`

```
  attr_reader :name, :location, :inventory

  def initialize(name, location)
    @name = name
    @location = location
    @inventory = []
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def inventory_count
+    @inventory_count = 0
+  end
```
kat3kasper, line 15:
```suggestion
  def inventory_count
    @inventory.count
  end
```

spec/car_spec.rb
```diff
@@ -37,4 +37,13 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  # it "has color" do
```
kat3kasper, line 41: Nice job on the TDD! The test looks correct

kat3kasper (commented):
Good start on this Ian. You struggled with setting up the initializers for the objects so that you could modify them later. I would review the following lessons so that you can get through creating objects a bit faster, scoping variables correctly and get through more iterations

https://backend.turing.edu/module1/lessons/scope

PR #64, opened 2021-12-08
Chris Kjolhede IC2

CKjolhede:


lib/car.rb
```diff
@@ -0,0 +1,27 @@
...
+    car_make = make_model.split[0]
+    p car_make
+
+  end
+
+  def model
+    make_model.split(" ")
+    car_model = make_model.split[1]
+    p car_model
+  end
```
kat3kasper, line 21: We don't want to print these values to the console
```suggestion
  def make
    make_model.split(" ")
    car_make = make_model.split[0]
  end

  def model
    make_model.split(" ")
    car_model = make_model.split[1]
  end
```

lib/car.rb
```diff
@@ -0,0 +1,27 @@
+class Car
```
kat3kasper, line 1: You are missing `.color` and the `paint!` method on the Car class

lib/dealership.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  end
+
+  def add_car(car)
+    @inventory.push(car)
+  end
+
+  def inventory
+    return @inventory
+  end
```
kat3kasper, line 18: You don't need these because you are using `attr_reader :inventory` which creates this method for us

lib/dealership.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def inventory_count
+    @inventory.count
+  end
```
kat3kasper, line 22: Nice work on utilizing @inventory to determine the count

spec/dealership_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    dealership.add_car(car_2)
+
+    expect(dealership.inventory_count).to eq(2)
+  end
+
+  it 'has an inventory' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+
+    expect(dealership.has_inventory?).to be(false)
+  end
```
kat3kasper, line 54: I'd probably also add a test for when `has_inventory?` is supposed to return `true`

kat3kasper (commented):
This is a pretty good start Chris. I would like to see more commits along the way of your coding. Make sure to slow down a bit and read through the specs. You missed the color and painting bit on Car. Keep practicing object creation so you can get a bit faster and get through more iterations.

PR #65, opened 2021-12-08
Tamara Car Dealership

tamara-builds:


spec/car_spec.rb
```diff
@@ -37,4 +37,17 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
```
dcoleman21, line 40: These tests look great.

lib/car.rb
```diff
@@ -0,0 +1,31 @@
...
+              :color
+
+  def initialize(type, monthly_payment, loan_length)
+    @type = type
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = nil
+  end
+
+  def make
```
dcoleman21, line 14: Nice job using the `.split` method here. Another way you could write this to shorten it up is
```
@type.split.first
```

lib/car.rb
```diff
@@ -0,0 +1,31 @@
...
+    make_split = @type.split(" ")
+    return make_split[0]
+  end
+
+  def model
```
dcoleman21, line 19: Nice job using the `.split` method here. Another way you could write this to shorten it up is
```
@type.split.last
```

lib/car.rb
```diff
@@ -0,0 +1,31 @@
...
+  attr_reader :type,
+              :monthly_payment,
+              :loan_length,
+              :color
+
+  def initialize(type, monthly_payment, loan_length)
+    @type = type
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = nil
```
dcoleman21, line 11: Nice job setting your `color` to `nil` here in the initialize.

lib/dealership.rb
```diff
@@ -0,0 +1,32 @@
+class Dealership
+  attr_reader :inventory,
+              :inventory_count
```
dcoleman21, line 3: You will want to have 
```
:name,
:location,
:inventory
```
in your `attr_reader`.  You will want to remove the `inventory_count` from your `attr_reader`.

lib/dealership.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize(name, location)
+    @name = name
+    @location = location
+    @inventory = []
+    @inventory_count = 0
```
dcoleman21, line 9: I would suggest making `inventory_count` a method instead of having it in your `initialize`. It isn't a counter, it is just returning the count of elements in your `inventory` array.
```
def inventory_count
   @inventory.length
end
```

lib/dealership.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def add_car(new_car)
+    @inventory << new_car
+    @inventory_count += 1
```
dcoleman21, line 14: See my comment about `inventory_count`. So this method only needs line 13

lib/dealership.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def has_inventory?
+    @inventory_count > 1
```
dcoleman21, line 18: So what is this method wanting to know? 
- whether or not there is anything in our `@inventory` array.

To have inventory, we would need our `@inventory` array to be greater than 0. So our `@inventory_count` would need to check that it is greater than 0.

lib/dealership.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def cars_by_make(make)
+    selected_cars = @inventory.select do |car|
+      car.type.split(' ')[0] == make
```
dcoleman21, line 23: Nice job using the `.select` enumerable here. 

Did you make a method called `make` in `car.rb` that returns the make? 
Can you just call that method on `car` and then see if it `== make`? So,
```
car.make == make
```

lib/dealership.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+  end
+
+  def total_value
+    inventory_value = @inventory.map do |car|
```
dcoleman21, line 28: Nice job using a `.map` here. 

So you are on the right track here by calling `car.total_cost` on line 29. You wouldn't put it into square brackets though. 

See how you created the `inventory_value` variable? You would just return `inventory_value` between lines 30 and 31. 

Since `.map` returns to us an array (which you are storing in the `inventory_value` variable you can then call `.sum` on that.  

```
inventory_value = @inventory.map do |car|
   car.total_cost
end
inventory_value.sum
```

dcoleman21 (commented):
Tamara, overall nice job writing your tests, paying attention to the interaction pattern details and then creating your methods. I have left you some feedback to review. Let me know if you have any questions.

lib/dealership.rb
tamara-builds: thank you very much for this.

PR #66, opened 2021-12-08
Alicia Watt

AliciaWatt:


spec/car_spec.rb
```diff
@@ -37,4 +40,12 @@
 
     expect(car.total_cost).to eq(54000)
   end
+
+  it "has color" do
+    expect(car.color).to eq([])
```
kat3kasper, line 45:
```suggestion
    expect(car.color).to eq(nil)
```

spec/car_spec.rb
```diff
@@ -37,4 +40,12 @@
...
+  end
+
+  it "is a color of paint" do
+  expect(car.paint!).to add(:blue)
+  end
```
kat3kasper, line 50: We want to pass a parameter to `.paint!`

```suggestion
  it "is a color of paint" do
    car.paint!("blue")
    expect(car.color).to eq("blue")
  end
```

lib/car.rb
```diff
@@ -0,0 +1,33 @@
+class Car
+  attr_reader :make,
+              :model,
+              :monthly_payment,
+              :loan_length
+              :color
+
+  def initialize(make_model, monthly_payment, loan_length)
+    @make = "Ford"
+    @model = "Mustang"
```
kat3kasper, line 10: We don't want to hard code these values in our initialize. Rather we want to set them via the data provided to us from the `make_model` variable:

@make = make_model.split.first
@model = make_model.split.last

lib/car.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def total_cost
+    loan_length * monthly_payment
+  end
+
+  def
+    color
+    @color = []
+  end
```
kat3kasper, line 27: Rathering than creating a method for color, we want to create a color attribute that we can have access to:

```
attr_reader :color

def initialize(make_model, monthly_payment, loan_length)
  ....
  @color = nil
end

```

then in our paint method. We can take in a color as a parameter and set our @color 

```
def paint!(color)
  @color = color
end
```

lib/car.rb
```diff
@@ -0,0 +1,33 @@
...
+    @make = "Ford"
+    @model = "Mustang"
+    @make_model = make_model
+    @monthly_payment = 1500
+    @loan_length = 36
+  end
+
+  def make_model
+   @make << model
+  end
```
kat3kasper, line 18: I don't think we need this.

lib/car.rb
```diff
@@ -0,0 +1,33 @@
...
+              :monthly_payment,
+              :loan_length
+              :color
+
+  def initialize(make_model, monthly_payment, loan_length)
+    @make = "Ford"
+    @model = "Mustang"
+    @make_model = make_model
+    @monthly_payment = 1500
+    @loan_length = 36
```
kat3kasper, line 13: We don't want to hard code these values either. We want to set them from the parameters being passed to `initialize`. Similar to what you did with `@make_model = make_model`

```suggestion
    @monthly_payment = monthly_payment
    @loan_length = loan_length
```

lib/dealership.rb
```diff
@@ -0,0 +1,20 @@
+class Dealership
+  attr_reader :name, :address, :inventory
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+  end
```
kat3kasper, line 8: This all looks good!

lib/dealership.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def inventory_count
+    inventory = 0
+  end
```
kat3kasper, line 12: Since we are storing all of our inventory in an array we can get the count of the inventory by calling the `.count` (or `.size` or `length they all work the same) method on the array. 

```suggestion
  def inventory_count
    @inventory.count
  end
```

lib/dealership.rb
```diff
@@ -0,0 +1,20 @@
...
+
+
+  def add_car(car)
+    @inventory << car
+    inventory.count
+  end
```
kat3kasper, line 18: Adding the car is right! Since we derive the inventory_count from the @inventory we only need the first line

```suggestion
  def add_car(car)
    @inventory << car
  end
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'rspec'
+require './lib/car'
+require './lib/dealership'
+
+RSpec.describe Dealership do
+  describe 'Iteration II' do
+  let(:car_1) {Car.new("Ford Mustang", 1500, 36)}
+  let(:car_2) {Car.new("Toyota Prius", 1000, 48)}
+  let (:dealership) {Dealership.new("Acme Auto", "123 Main Street")}
```
kat3kasper, line 9: I like the use of `let` up here so that you could reuse the car_1, car_2, and dealership objects across multiple tests.

spec/dealership_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+it 'counts inventory' do
+expect(dealership.inventory_count).to eq(0)
+end
+
+it 'adds cars' do
+  dealership.add_car(car_1)
+  dealership.add_car(car_2)
+  expect(dealership.inventory).to eq([car_1, car_2])
+  expect(dealership.inventory_count).to eq(2)
+  end
```
kat3kasper, line 36: Great job on these tests!

kat3kasper (commented):
Alicia you seem to be struggling quite a bit with object creation. I'd review the following lessons and perhaps reach out to a mentor or practice on some of your old assignments

https://backend.turing.edu/module1/lessons/methods_and_return_values

PR #67, opened 2021-12-08
Max Wagner car_dealership

MWagner3:


lib/dealership.rb
```diff
@@ -0,0 +1,37 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def cars_by_make(make)
+    same_make = []
+    @inventory.each do |inventory|
+      if inventory.car.make == make
```
cjsim89, line 28: This should read:
`if inventory.make == make` instead of `inventory.car.make`, because in this scope your block variable is `inventory`, which maybe to read better should just be `car`, since your inventory is an array of Cars. 

This method throws an error where Ruby is saying, "There is no "car" on `inventory`", because "inventory" is actually the car.

lib/dealership.rb
```diff
@@ -0,0 +1,37 @@
...
+      @same_make << @inventory.car
```
cjsim89, line 29: Careful of your `@` character here - `same_make` is not a class-level variable, it is a method variable that you created on line 26, and so does not need the @. Same with `@inventory` - this is inside of your .each, so you want to add the actual `inventory` block variable (which again is technically a Car since we're looping through) to this array.

lib/dealership.rb
```diff
@@ -0,0 +1,37 @@
...
+
+     end
+     return same_make
```
cjsim89, line 32: This `return` is placed inside of your `each` block - it should be moved down to after line 33. Right now it returns far too early, or not at all.

spec/car_spec.rb
```diff
@@ -3,38 +3,44 @@
 
 RSpec.describe Car do
   it 'exists' do
-    car = Car.new("Ford Mustang", 1500, 36)
```
cjsim89, line 6: Unfortunately this was the right approach - there should only be 3 arguments in your `car#initialize` method. Separating your first arg into 2 goes against the interaction pattern where all the cars have both make & model in one string. The expectation is to then separate them by the space via `.split`.

spec/car_spec.rb
```diff
@@ -3,38 +3,44 @@
...
 
     expect(car.loan_length).to eq(36)
   end
 
   it 'has a total cost' do
-    car = Car.new("Ford Mustang", 1500, 36)
+    car = Car.new("Ford", "Mustang", 1500, 36)
 
-    expect(car.total_cost).to eq(54000)
+    expect(car.total_cost(54000)).to eq(54000)
```
cjsim89, line 38: `#total_cost` should not take an argument, it should only return a result. The method should perform a calculation based on monthly cost and months (cost * months). Looks like your first approach was closer here.

cjsim89: This is what the interaction pattern said: 
```
pry(main)> car.total_cost
#=> 54000
```

Notice that `car.total_cost` does not have `()` after it - so it won't take an argument. It only returns a result.

lib/car.rb
```diff
@@ -0,0 +1,24 @@
+class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length, :total_cost
+  def initialize(make, model, monthly_payment, loan_length)
```
cjsim89, line 3: The interaction pattern shows 3 arguments, you have 4 -- it is very important that you follow the patterns exactly to the letter.

cjsim89: Note that you CAN actually perform logic inside of your `initialize` methods - the idea here is to take one argument of both make and model, and then split them via `.split(" ")`, a Ruby method that can separate parts of a string into an array.

PR #54, opened 2021-12-08
Conor Barthel

conorbarthel:

PR #30, opened 2021-10-20
feat add IT 1 to 3

Eagerlearn:
Craig Cardwell

PR #68, opened 2021-12-08
Jimmy Page

jpage13:
finished iteration 4

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,14 @@
+class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length
+
+  def initialize(make_and_model, monthly_payment, loan_length)
+    @make = make_and_model.split.first
+    @model = make_and_model.split.last
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+  end
+
+  def total_cost
+    @monthly_payment * @loan_length
+  end
+end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,14 +1,19 @@
 class Car
-  attr_reader :make, :model, :monthly_payment, :loan_length
+  attr_reader :make, :model, :monthly_payment, :loan_length, :color
 
   def initialize(make_and_model, monthly_payment, loan_length)
     @make = make_and_model.split.first
     @model = make_and_model.split.last
     @monthly_payment = monthly_payment
     @loan_length = loan_length
+    @color = nil
   end
 
   def total_cost
     @monthly_payment * @loan_length
+  end
+
+  def paint!(paint_color)
+    @color = paint_color
   end
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,0 +1,1 @@
+class Dealership
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -33,4 +33,8 @@ class Dealership
   def total_value
     total_value_helper.sum
   end
+
+  def details
+    details_hash = Hash["total_value"=>total_value, "address"=> address]
+  end
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -1,1 +1,8 @@
 class Dealership
+  attr_reader :name, :address
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+  end
+end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -13,5 +13,11 @@ class Dealership
     @inventory_count += 1
   end
 
+  def has_inventory?
+    @inventory_count > 0
+  end
+
+  def cars_by_make(arg)
+  end 
 
 end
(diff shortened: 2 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,19 @@
+class Car
```
kat3kasper, line 1: This all looks good! Clean and easy to read

lib/dealership.rb
```diff
@@ -0,0 +1,72 @@
+class Dealership
+  attr_reader :name, :address, :inventory, :inventory_count, :dealership_value
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+    @inventory_count = 0
+    @dealership_value = 0
```
kat3kasper, line 9: I don't think you need `dealership_value` as an instance variable rather we could calculate this in a method when necessary. Also I don't think you are using it anywhere

lib/dealership.rb
```diff
@@ -0,0 +1,72 @@
...
+    @inventory << car
+    @inventory_count += 1
+  end
+
+  def has_inventory?
+    @inventory_count > 0
+  end
+
+  def cars_by_make(chosen_make)
+    @inventory.find_all do |car|
```
kat3kasper, line 22: Like the use of `find_all` 💪

lib/dealership.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def total_value
+    total_value_helper.sum
+  end
+
+  def details
+    details_hash = Hash["total_value"=>total_value, "address"=> address]
+  end
+
+  def average_price_of_car
```
kat3kasper, line 41: Nice use of reusing `total_value` and `inventory_count` to calculate the average price of the car!

spec/dealership_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+    expect(dealership.cars_sorted_by_price).to eq([car_1, car_2, car_3, car_4])
```
kat3kasper, line 117: So what you expect isn't exactly correct

    car_1 = Car.new("Ford Mustang", 1500, 36) #54,000
    car_2 = Car.new("Toyota Prius", 1000, 48)  #48,000
    car_3 = Car.new("Toyota Tercel", 500, 48)  #24,000
    car_4 = Car.new("Chevrolet Bronco", 1250, 24) #30,000

so the result from cars_sorted_by_price should be `[car_3, car_4, car_2, car_1]`

lib/dealership.rb
```diff
@@ -0,0 +1,72 @@
...
+    total_value/ inventory_count
+  end
+
+  def cars_sorted_by_price
+    #they already are
+    @inventory
```
kat3kasper, line 47: Unfortunately they are not. Check out my comments in your tests below

lib/dealership.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+  end
+
+  def chevy_cars_helper
+    @inventory.select do |car|
+      car.make == "Chevrolet"
+    end
+  end
+
+  def inventory_hash
```
kat3kasper, line 68: So we want this to be able to work no matter what the make of the car and not only for Ford, Toyota, and Chevrolet so we don't want to be hard coding the values here. One way to do that could be the following:

```
  def inventory_hash
    inventory_hash = {}
    @inventory.each do |car|
      if inventory_hash[car.make]
        inventory_hash[car.make] << car
      else
      end
    end
    inventory_hash
  end
```

spec/dealership_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    expect(dealership.cars_by_make("Toyota")).to eq([car_2, car_3])
+    expect(dealership.cars_by_make("Ford")).to eq([car_1])
+  end
+
+  it 'can tell you the total value of cars in the dealership' do
+    dealership = Dealership.new("Acme Auto", "123 Main Street")
+    car_1 = Car.new("Ford Mustang", 1500, 36)
+    car_2 = Car.new("Toyota Prius", 1000, 48)
+    car_3 = Car.new("Toyota Tercel", 500, 48)
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
```
kat3kasper, line 71: Can you think of a way to use `before(:each)` or `let` so that you don't need to redefine the dealership and cars for every test?

kat3kasper (commented):
Great start Jimmy! 🎉 A bit of hard coding in there which isn't really what we want. Keep up the good work with the commit history!!

PR #69, opened 2021-12-08
Steven James

stevenjames-turing:
Iteration 1-3 complete. Parts of I-4 complete.

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,11 @@
+class Car
+
+  def initialize(type, monthly_payment, loan_length)
+    @type =
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+  end
+
+  
+
+end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,11 +1,18 @@
 class Car
+  attr_reader :make, :model, :monthly_payment, :loan_length
 
   def initialize(type, monthly_payment, loan_length)
-    @type =
+    @type = type
+    @make = @type.split(" ").first
+    @model = @type.split(" ").last
     @monthly_payment = monthly_payment
     @loan_length = loan_length
   end
 
-  
+  def total_cost
+    @monthly_payment * @loan_length
+  end
+
+
 
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,5 +1,5 @@
 class Car
-  attr_reader :make, :model, :monthly_payment, :loan_length
+  attr_reader :make, :model, :monthly_payment, :loan_length, :color
 
   def initialize(type, monthly_payment, loan_length)
     @type = type
@@ -7,12 +7,15 @@ class Car
     @model = @type.split(" ").last
     @monthly_payment = monthly_payment
     @loan_length = loan_length
+    @color = nil
   end
 
   def total_cost
     @monthly_payment * @loan_length
   end
 
-
+  def paint!(color)
+    @color = color
+  end
 
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -38,12 +38,15 @@ class Dealership
   end
 
   def details
-    details = {
-
-    }
+    details = {}
     details["total_value"] = total_value
     details["address"] = @address
-    details 
+    details
+  end
+
+  def average_price_of_car
+    average_price = total_value / inventory_count
+    average_price.to_s.insert(-4, ",")
   end
 
 end
--- a/lib/dealership.rb
+++ b/lib/dealership.rb
@@ -50,13 +50,9 @@ class Dealership
   end
 
   def cars_sorted_by_price
-    prices = []
-    inventory.each do |car|
-      prices << @price
+    inventory.sort_by do |car|
+      car.total_cost
     end
-
-
-
   end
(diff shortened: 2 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,21 @@
...
+  attr_reader :make, :model, :monthly_payment, :loan_length, :color
+
+  def initialize(type, monthly_payment, loan_length)
+    @type = type
+    @make = @type.split(" ").first
+    @model = @type.split(" ").last
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = nil
+  end
```
rtillies, line 11: Good work using `split` on your `make_model` argument. Can you think of a different way to implement this to prevent having to call `split` multiple times within `initialize`?

lib/dealership.rb
```diff
@@ -0,0 +1,65 @@
...
+    details = {}
+    details["total_value"] = total_value
+    details["address"] = @address
+    details
+  end
+
+  def average_price_of_car
+    average_price = total_value / inventory_count
+    average_price.to_s.insert(-4, ",")
+  end
```
rtillies, line 50: Perfect! Great job figuring this out!

lib/dealership.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def cars_sorted_by_price
+    inventory.sort_by do |car|
+      car.total_cost
+    end
+  end
```
rtillies, line 56: Nice use of the `sort_by` enumerable. This is covered in Week 5, great job learning about it early.

lib/dealership.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def inventory_hash
+    inventory_hash = {}
+    inventory.each do |car|
+      inventory_hash[car.make] = car
+    end
+    inventory_hash
+  end
```
rtillies, line 64: This is very close. Can you figure out why this does not work as described in the interaction pattern?

spec/dealership_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+    car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+
+    dealership.add_car(car_1)
+    dealership.add_car(car_2)
+    dealership.add_car(car_3)
+    dealership.add_car(car_4)
+
+    expect(dealership.inventory_hash.keys).to eq(["Ford", "Toyota", "Chevrolet"])
+
+  end
```
rtillies, line 146: This tests passes as written, but it does not match the interaction pattern.
```
pry(main)> dealership.inventory_hash
#=> {"Ford"=> [#<Car:0x00007fccd29b5720...>],
"Toyota"=> [#<Car:0x00007fccd2985f48...>, #<Car:0x00007fccd383c2d0...>],
"Chevrolet"=> [#<Car:0x00007fccd297dc30...>]}
```

lib/dealership.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def cars_by_make(make)
+    cars_by_make = []
+    @inventory.each do |car|
+      if car.make == make
+        cars_by_make << car
+      end
+    end
+    cars_by_make
+  end
```
rtillies, line 30: Great use of `each` and an accumulator array to make this method work.

lib/dealership.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def total_value
+    total_value = 0
+    @inventory.each do |car|
+      total_value += car.total_cost
+    end
+    total_value
+  end
```
rtillies, line 38: Great use of `each` and an accumulator variable to make this method work.

rtillies (commented):
Congrats on nearly completing Iteration 4! I can tell you are using TDD to write good tests and allowing the tests to write quality code. Most of my "critical" comments are nit-picky, but the goal for you is small incremental changes, even at this early stage of your development career. Good work on making 15 commits. Keep learning and keep coding!

rtillies:

lib/dealership.rb
stevenjames-turing: When `.each` is iterating through `inventory`, it is assigning the current car as the value for the `car.make` key inside of the hash. However, if that key already has a value assigned it is re-assigning the key to the new value instead of adding it to the key in an array. This is making the return value of the method only return the last car to be assigned to each unique key instead of all of the cars that were added to the keys.  If I had more time, I think I could have figured this one out, but that was the hang-up I kept running into.

spec/dealership_spec.rb
stevenjames-turing: For future TDDing, how would you write the test if you wanted it to return an array of class instances? I know I had the issue in this code block with the values of each key overriding each other and that I would need to write another test that showed Toyota having 2 values assigned, but I actually didn't know how outside of something along the lines of `dealership.inventory_hash["Toyota"].count`. Would something like that work for matching the interaction pattern?

lib/dealership.rb
stevenjames-turing: I took some time on Friday and ended up figuring this problem out. I've pushed new changes of it to Github if you're interested in checking it out.

PR #70, opened 2021-12-08
Michael Hicks

michaeljhicks:
I was freaking out about mistakenly doing the pull request and overwriting the main....sounds really stupid, but I wanted to 100% sure.

### DIFF
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,5 +1,7 @@
 require 'rspec'
 require './lib/car'
+require './lib/dealership'
+require 'pry'
 
 class Car
   attr_reader :make_and_model,
@@ -27,6 +29,11 @@ class Car
 
   def total_cost
     total_cost = @loan_length * @monthly_payment
+
+  end
+
+  def paint!(color)
+    @color = color
 
   end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -25,4 +25,8 @@ class Car
 
   end
 
+  def total_cost
+    total_cost = @loan_length * @monthly_payment
+
+  end
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -19,6 +19,10 @@ class Car
 
   end
 
+  def model
+    make_and_model_split = make_and_model.split()
+    model = make_and_model_split[1]
 
+  end
 
 end
--- a/lib/car.rb
+++ b/lib/car.rb
@@ -1,0 +1,24 @@
+require 'rspec'
+require './lib/car'
+
+class Car
+  attr_reader :make_and_model,
+              :monthly_payment,
+              :loan_length,
+              :color
+  def initialize(make_and_model, monthly_payment, loan_length)
+    @make_and_model = make_and_model
+    @monthly_payment = monthly_payment
+    @loan_length = loan_length
+    @color = nil
+  end
+
+  def make
+    make_and_model_split = make_and_model.split()
+    make = make_and_model_split[0]
+
+  end
+
+
+
+end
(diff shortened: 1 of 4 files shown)


lib/car.rb
```diff
@@ -0,0 +1,39 @@
...
+    make_and_model_split = make_and_model.split()
+    make = make_and_model_split[0]
+
+  end
+
+  def model
+    make_and_model_split = make_and_model.split()
+    model = make_and_model_split[1]
+
+  end
```
rtillies, line 28: This works just fine. Can you come up with a different way to make this work instead of creating two separate methods that are nearly clones of each other?

lib/dealership.rb
```diff
@@ -0,0 +1,67 @@
...
+    @inventory.push car
+    @total_value += car.total_cost
+  end
+
+  def has_inventory?
+    if @inventory.count != 0
+      p true
+    else
+      p false
+    end
```
rtillies, line 30: Honestly, I'm surprised this works. Do you want this method to print `true`/`false` to the console, or return `true`/`false` to the method calling it?

lib/dealership.rb
```diff
@@ -0,0 +1,67 @@
...
+require 'pry'
+
+class Dealership
+  attr_reader :name, :address, :inventory, :inventory_count, :total_value
+
+  def initialize(name, address)
+    @name = name
+    @address = address
+    @inventory = []
+    @total_value = 0
```
rtillies, line 13: I would generally avoid creating a `@total_value` attribute, because it is dependent on the contents of the `@inventory` array attribute. My general rule is that if an attribute depends on another attribute, it shouldn't be its own attribute. Instead, write a method that calculates `total_value`.

lib/dealership.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def inventory_count
+    @inventory.count
+  end
+
+  def add_car(car)
+    @inventory.push car
+    @total_value += car.total_cost
+  end
```
rtillies, line 23: In an upcoming class, you'll learn about SRP: Single Responsibility Principle. In part, it states that a method should be responsible for a single action. If a method does two (or more) things, it should be separated into two (or more) methods. In this case, your `add_car` method is responsible for adding the car to the `@inventory` array AND updating the `@total_value`, which would be a violation of SRP. A solution would be for `add_car` to only add the car to the array, and have a separate method that calculates `total_value` (which is related to my comments above about having a `@total_value` attribute).

lib/dealership.rb
```diff
@@ -0,0 +1,67 @@
...
+      "total_value" => @total_value,
+      "address" => @address
+    }
+    return details_hash
+  end
+
+  def average_price_of_car
+    avg_price = @total_value / (@inventory.count)
+    # return avg_price.to_s
+  end
```
rtillies, line 59: You are on the right track with the `to_s` method. What else do you need to make this work to match the interaction pattern?

lib/dealership.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def cars_sorted_by_price
+    sorted_cars = []
+    
+
+  end
```
rtillies, line 65: I imagine you just ran out of time here (?)

rtillies (commented):
Congrats on completing Iteration 3. I can tell you are using TDD to write good tests and allowing the tests to write quality code. I fully expect that you would have completed Iteration 4 with more time. Good work on making 8 commits; make sure you commit after every win (e.g. after every successful test). Keep learning and keep coding!

rtillies:

PR #71, opened 2021-12-08
Jacob Jenner

jennerj:
I messed up a lot here, like I said I am going to take some time today to rename everything so I wont run into this again.

### DIFF
--- a/car_dealership_2111/lib/car.rb
+++ b/car_dealership_2111/lib/car.rb
@@ -1,0 +1,24 @@
+class Car
+  attr_reader :make,
+              :model,
+              :monthly_payment,
+              :loan_length,
+              :total_cost,
+              :color
+
+  def initialize(make_model, monthly_payment, loan_length)
+    @make            = make_model.split.first
+    @model           = make_model.split.last
+    @monthly_payment = monthly_payment
+    @loan_length     = loan_length
+    @color           = nil
+  end
+
+  def paint!(color)
+    @color = color
+  end
+
+  def total_cost
+    @loan_length * @monthly_payment
+  end
+end
--- a/car_dealership_2111/lib/dealership.rb
+++ b/car_dealership_2111/lib/dealership.rb
@@ -1,0 +1,56 @@
+class Dealership
+  attr_reader :name,
+              :address,
+              :inventory,
+              :inventory_count
+
+  def initialize(name, address)
+    @name            = name
+    @address         = address
+    @inventory       = []
+    @inventory_count = 0
+  end
+
+  def add_car(car)
+    @inventory << car
+    @inventory_count += 1
+  end
+
+  def has_inventory?
+    @inventory_count > 0
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |car|
+      car.make == make
+    end
+  end
+
+  def total_value
+    @inventory.sum do |car|
+      car.total_cost
+    end
+  end
+
+  def details
+    {
+    "total_value" => total_value,
+    "address" => @address
+    }
+  end
+
+  def average_price_of_car
+    average = @total_value / @inventory_count
+    average.to_s.insert(2, ",")
+
+  end
+
+  def cars_sorted_by_price
+    @inventory.sort_by do |car|
+      car.total_cost
+    end
+  end
+
+  def inventory_hash
+  end
+end
--- a/car_dealership_2111/spec/dealership_spec.rb
+++ b/car_dealership_2111/spec/dealership_spec.rb
@@ -1,0 +1,100 @@
+require './lib/car'
+require './lib/dealership'
+
+RSpec.describe Dealership do
+  before(:each) do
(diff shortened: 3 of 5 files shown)


car_dealership_2111/lib/dealership.rb
```diff
@@ -0,0 +1,56 @@
...
+  attr_reader :name,
+              :address,
+              :inventory,
+              :inventory_count
+
+  def initialize(name, address)
+    @name            = name
+    @address         = address
+    @inventory       = []
+    @inventory_count = 0
```
kat3kasper, line 11: Another way you could have done `inventory_count` would be to calculate it based on `@inventory`:

```
def inventory_count
  @inventory.length
end
```

Then when adding the car you wouldn't need to update the size.

car_dealership_2111/lib/dealership.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def cars_by_make(make)
+    @inventory.select do |car|
+      car.make == make
+    end
+  end
+
+  def total_value
+    @inventory.sum do |car|
```
kat3kasper, line 30: Love the use of the `sum` and `select` enumerables! 😍

car_dealership_2111/lib/dealership.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def details
+    {
+    "total_value" => total_value,
+    "address" => @address
+    }
+  end
+
+  def average_price_of_car
+    average = @total_value / @inventory_count
```
kat3kasper, line 43: You were really close here! total_value is a method rather than an instance var

```suggestion
    average = total_value / @inventory_count
```

car_dealership_2111/lib/car.rb
```diff
@@ -0,0 +1,24 @@
...
+              :total_cost,
+              :color
+
+  def initialize(make_model, monthly_payment, loan_length)
+    @make            = make_model.split.first
+    @model           = make_model.split.last
+    @monthly_payment = monthly_payment
+    @loan_length     = loan_length
+    @color           = nil
+  end
```
kat3kasper, line 15: Great setup!

car_dealership_2111/lib/car.rb
```diff
@@ -0,0 +1,24 @@
+class Car
+  attr_reader :make,
```
kat3kasper, line 2: Nice job using `attr_reader` for these attributes!

car_dealership_2111/spec/dealership_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  it 'has inventory' do
+    expect(@dealership.has_inventory?).to eq(false)
+
+    @dealership.add_car(@car_1)
+    @dealership.add_car(@car_2)
+    @dealership.add_car(@car_3)
+    @dealership.add_car(@car_4)
+
+    expect(@dealership.has_inventory?).to eq(true)
+  end
```
kat3kasper, line 41: I like that you checked `has_inventory?` before and after 👍

car_dealership_2111/spec/dealership_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  it 'returns inventory sorted in a hash' do
+    @dealership.add_car(@car_1)
+    @dealership.add_car(@car_2)
+    @dealership.add_car(@car_3)
+    @dealership.add_car(@car_4)
+
+    expect(@dealership.inventory_hash).to be_a({})
+    expect(@dealership.inventory_hash.keys).to eq(["Ford", "Toyota", "Chevrolet"])
+    expect(@dealership.inventory_hash.values).to eq([@car_1, @car_2, @car_3, @car_4])
```
kat3kasper, line 97: Not exactly what you were looking for test wise, but I'm glad you used TDD to write the test before trying to implement `inventory_hash`

car_dealership_2111/spec/dealership_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+require './lib/dealership'
+
+RSpec.describe Dealership do
+  before(:each) do
+    @dealership = Dealership.new("Acme Auto", "123 Main Street")
+    @car_1 = Car.new("Ford Mustang", 1500, 36)
+    @car_2 = Car.new("Toyota Prius", 1000, 48)
+    @car_3 = Car.new("Toyota Tercel", 500, 48)
+    @car_4 = Car.new("Chevrolet Bronco", 1250, 24)
+  end
```
kat3kasper, line 11: Nice use of before(:each)

kat3kasper (commented):
Nice work Jacob 💪 !! Would like to see more of a commit history along the way. It would be a good idea to commit after each iteration. I ran into a bit of trouble because you seemed to have moved all of your code into another `car_dealership_2111` directory. I moved all the files to the right place and was able to run the tests.

